PR #400, opened 2026-05-19
feat(core): populate git.* properties on compiled entries

stasson:
## Summary

Fills the four `entry.properties.git` fields (`createdAt`, `modifiedAt`, `contributors`, `revision`) that were declared on the model but never populated. Mirrors the `file.*` slice (PR #391).

- **core stays Node-safe**: new optional `CompileOptions.gitFile` callback (analogous to `statFile`) keeps all git I/O out of `core/`. `resolveGit()` centralises the PII gate and makes `contributors` deterministic (dedup + sort). When the callback is absent or returns `undefined`, `properties.git` stays unset.
- **CLI wiring**: real `gitFile` via `git log --follow` wired into the shared `compileProject` helper (covers `compile`, `export`, `show`, `context`, `dependents`, `report`, `insert`, `create`, `next-id`, `lint`, `doc`) and the `book build` path. `markspec compile` gains `--with-contributors` (default off — `git.contributors` is PII-adjacent per ADR-006; a published `/api/` is world-readable, per `docs/specs/markspec-compile-output.md` section 6).
- **Graceful degradation**: `git` not on PATH, untracked file, or no run permission yields `undefined`, never throws.
- **createdAt/modifiedAt semantics**: `createdAt` is the oldest commit author date, `modifiedAt` the newest. `git log` lists newest-first, so these map to the last/first log line respectively. The handoff's "first = createdAt" shorthand assumed `--reverse` ordering; the implementation follows the model's documented intent (first commit gives `createdAt`).

No doc change: `docs/specs/markspec-compile-output.md` section 6 already specs `--with-contributors` and `git.*`; this PR makes the spec true (same precedent as PR #391).

## Test Plan

- [x] Unit (`core/compiler/mod_test.ts`): populate from mock callback; absent without callback; callback returns `undefined` -> git absent, no crash; `withContributors:true` -> deduped + sorted; default false strips contributors even when callback returns them.
- [x] E2E (`tests/e2e/compile_git_properties_test.ts`): graceful degradation (non-git dir, with/without run perm, `--output`); real-git present case (`compile --format json`, `--with-contributors`, `--output` compiled.json).
- [x] `just build` green; `deno fmt --check` + `dprint check` clean; full suite 1579 passed / 0 failed.

### DIFF
--- a/packages/markspec/cli/commands/book.ts
+++ b/packages/markspec/cli/commands/book.ts
@@ -9,6 +9,7 @@ import { Command } from "@cliffy/command";
 import type { BookStructure, Chapter } from "../../book/mod.ts";
 import {
   loadActiveProfile,
+  makeGitFile,
   notImplemented,
   requireProjectConfig,
 } from "../helpers.ts";
@@ -58,6 +59,7 @@ export const bookCmd = new Command()
       profile: bookChain?.effective ?? undefined,
       statFile: (p) =>
         Deno.stat(p).then((s) => ({ mtime: s.mtime })).catch(() => undefined),
+      gitFile: makeGitFile(false),
     });
 
     const result = buildBook(structure, {
--- a/packages/markspec/cli/commands/compile.ts
+++ b/packages/markspec/cli/commands/compile.ts
@@ -22,13 +22,24 @@ export const compileCmd = new Command()
     "Entry count at which to switch to NDJSON streaming output",
     { default: 1000 },
   )
+  .option(
+    "--with-contributors",
+    "Include git contributor names in properties.git (PII-adjacent, ADR-006; off by default)",
+  )
   .arguments("<paths...:string>")
   .action(
     async (
-      _options: { format?: string; output?: string; splitThreshold: number },
+      _options: {
+        format?: string;
+        output?: string;
+        splitThreshold: number;
+        withContributors?: boolean;
+      },
       ...paths: string[]
     ) => {
-      const { result, chain } = await compileProject(paths);
+      const { result, chain } = await compileProject(paths, {
+        withContributors: _options.withContributors,
+      });
 
       if (_options.output) {
         const {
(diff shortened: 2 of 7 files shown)


stasson:
## Code review (automated, post-PR per AGENTS.md)

**Verdict: Approve** — correct, follows the `file.*` precedent, well-tested, secure. Three non-blocking notes.

### Overview
Populates `entry.properties.git` (createdAt/modifiedAt/contributors/revision) via an injected `CompileOptions.gitFile` callback so `core/` stays Node-safe; CLI supplies a `git log --follow` implementation. `--with-contributors` gates the PII-adjacent field (default off).

### Correctness
- `git log` is newest-first by default, so `modifiedAt = dates[0]` / `createdAt = dates[last]` is the right mapping (oldest commit → createdAt). Matches the model's documented intent.
- `--` before the pathspec and argv-array `Deno.Command` (no shell) — no path-injection surface. PII default-off. Sound.
- `resolveGit` empty-object guard prevents emitting `git: {}`; contributors dedup+sort is deterministic.
- Graceful degradation verified by E2E for the no-`git`, untracked, and no-run-permission paths.

### Non-blocking notes
1. **`%h` short SHA is repo-size-dependent.** Git auto-sizes abbreviated SHAs by object count, so `revision` length (hence compiled-output bytes) can differ across environments/time. The handoff and model both say "short SHA" so this is likely intentional, but it slightly weakens the "deterministic output" guarantee for `compile`/`export`. Worth a conscious confirm; `%H` would be strictly reproducible.
2. **Per-file git spawn cost.** 2 (or 3 with `--with-contributors`) `git log --follow` processes per source file. Fine at current scale and consistent with the per-file callback design, but `compile` over a large tree scales with file-count × process-spawn. A future single-pass `git log --name-only` batch could amortize this.
3. **`gitLines` catch-all** maps every failure (incl. a corrupt repo) to "no git data". Intended per the never-throw requirement; a one-line debug diagnostic could aid future triage without changing behavior.

### Tests
Unit RED→GREEN for populate / absent / undefined / contributor strip / dedup-sort; E2E for graceful degradation + real-repo present case + `--output`. `just build` green, `deno fmt --check` + `dprint check` clean, full suite 1579/0.

PR #405, opened 2026-05-22
feat(windows): native Windows support across CLI, LSP, and installer

### LINKED ISSUE #403: Epic: Windows support — CLI, VSCode extension, installer
## Problem

MarkSpec is published as a cross-platform tool — `deno compile` produces a Windows binary, and the VSCode extension ships a `win32-x64` VSIX — but the codebase has never been audited or tested on native Windows. WSL is the only path that has been exercised. Several latent bugs and gaps prevent Windows users from running the toolchain natively today.

### Broken on Windows

- **LSP file URIs are malformed.** [`packages/markspec/lsp/util.ts:22`](packages/markspec/lsp/util.ts#L22) uses `new URL("file://" + path)`. On Windows this produces `file://C:\foo\bar` instead of the spec-required `file:///C:/foo/bar`. Breaks diagnostics, go-to-definition, references, rename, hover, document symbols. Same shape in [`packages/markspec/cli/commands/doc.ts:25`](packages/markspec/cli/commands/doc.ts#L25).
- **Installer scripts reject Windows.** [`bootstrap`](bootstrap) and [`install.sh`](install.sh) are bash-only. `bootstrap` explicitly errors with "unsupported OS (use WSL on Windows)". Native Windows users have no installation path.

### Fragile on Windows

- **Hardcoded path separators.** [`packages/markspec/lsp/server.ts:737`](packages/markspec/lsp/server.ts#L737) (`walkDirectory`) and [`packages/markspec/mcp/project.ts:157`](packages/markspec/mcp/project.ts#L157) (`walkFs`) construct paths with `${dir}/${entry.name}`. Deno's `readDir` tolerates `/`, but mixed separators leak into `SourceLocation.file` and compound with the URI bug above.
- **Line endings.** `.split("\n")` is used throughout the parser, formatter, and LSP. `.gitattributes` does not normalize `.md` or `.ts`. CRLF checkouts leave stray `\r` on every line. No test exercises CRLF input.

### Untested

- The compiled `markspec.exe` is built in CI ([`.github/workflows/release.yaml:29`](.github/workflows/release.yaml#L29)) but **no Windows runner executes the test suite**. "Compiles" is not "runs correctly".
- The VSCode extension is correctly platform-aware in [`editors/vscode/src/serverOptions.ts:65`](editors/vscode/src/serverOptions.ts#L65) and [`editors/vscode/src/mcpDefinition.ts:89`](editors/vscode/src/mcpDefinition.ts#L89) (resolves `markspec.exe`), but the end-to-end flow has not been exercised on Windows.

## Scope

This epic covers full Windows support across three surfaces:

1. **CLI** — `markspec.exe` runs all subcommands correctly on Windows (PowerShell, cmd, Git Bash) against CRLF and LF checkouts.
2. **VSCode extension** — LSP and MCP integration work end-to-end on Windows, with correct file URIs and path handling.
3. **Installer** — Native Windows installation path (PowerShell script, Scoop, or winget — to be decided in the spec).

## Acceptance criteria

- All e2e tests pass on a Windows runner in CI.
- `pathToUri` round-trips correctly for Windows paths (`C:\foo\bar` ↔ `file:///C:/foo/bar`).
- Parser and formatter handle CRLF input without leaking `\r` into entry bodies.
- An installation method exists for native Windows users (not WSL).
- VSCode extension verified working on Windows 11 with a real MarkSpec project.
- `docs/guide/` documents the Windows installation and known caveats.

## Out of scope

- WSL improvements (WSL already works as a Linux environment).
- ARM Windows (`aarch64-pc-windows-msvc`) — defer until x86_64 is solid.
- Windows-specific UI features in the VSCode extension.

## Next step

Spec to be authored under `docs/product/` and ADR(s) under `docs/architecture/` covering path handling, line-ending policy, and installer choice. Stories will be split out of this epic once the spec is reviewed.

stasson:
## Summary

Native Windows support across the CLI, VSCode extension, and installer. Implements all 7 stories from [`docs/product/windows-support.md`](docs/product/windows-support.md). Closes #403.

## What changed

| STK | Area | Change |
|---|---|---|
| WIN-0002 | LSP / CLI | `pathToUri` & `uriToPath` use `@std/path`'s platform-aware `toFileUrl` / `fromFileUrl`; same fix in `cli/commands/doc.ts` and test helpers. Windows drive-letter paths now round-trip correctly. |
| WIN-0003 | core / LSP / MCP / CLI | `walkDirectory`, `walkFs`, and other path builders use `@std/path` `join` instead of `${dir}/${name}`. |
| WIN-0004 | parser / formatter | CRLF normalised at parse + AST-build boundaries; formatter detects source convention and restores it on write-back. AST-fidelity matrix stays at **0/58** because `buildBodyAst` normalises before mdast parses. |
| WIN-0001 / WIN-0007 | CI | `ci.yaml` `test` job now matrixes `ubuntu-latest`, `windows-latest`, `macos-latest`. `fail-fast: false`. |
| WIN-0006 | installer | New `install.ps1` mirroring `install.sh`. New `pwsh-installer` CI job parses the script on `windows-latest`. End-to-end install-from-release smoke test deferred to a follow-up in `release.yaml` (chicken-and-egg on PR branches). |
| WIN-0005 | VSCode | New `editors/vscode/README.md` with a 13-row manual Windows smoke-test checklist + CRLF and path-handling spot checks. |
| WIN-0008 | docs | `docs/guide/installation.md` documents the Windows PowerShell path with the SmartScreen / signing caveat. `README.md` links to it. `bootstrap` keeps WSL guidance for contributors and now also points end users at `install.ps1`. |

## Test plan

- [x] `just check` — 1648 pass / 0 fail / 1 ignored on macOS (1645 baseline + 3 CRLF e2e + 14 unit + 7 Windows-path unit)
- [x] `just ast-fidelity-matrix` — surface 0/58 maintained
- [x] `dprint check` clean
- [x] `deno fmt --check` clean
- [x] `deno lint` clean
- [ ] **CI on this PR** — first run of `Test (windows-latest)` will validate the changes against a real Windows runner. Any regression there is the new gate working as intended.
- [ ] **Manual smoke test on Windows 11** — checklist in `editors/vscode/README.md` to be run before the next release.

## Follow-ups (not in this PR)

- **Branch protection.** Repo admin should add `Test (windows-latest)` to the required-checks list on `main`.
- **Code signing.** SmartScreen warning on first run is currently expected; tracking via `## Open questions` in the spec.
- **Install-from-release smoke test.** Add a job to `release.yaml` that runs `install.ps1` against the just-published release and asserts `markspec --version`.
- **Scoop / winget manifests.** Deferred per the spec's installer decision.
- **e2e helper hardening.** The existing test helpers (`tests/e2e/helpers.ts` etc.) still build temp-dir paths with `${dir}/${name}` — Deno tolerates `/` on Windows so they work, but they're not idiomatic. Tracked separately if it surfaces.

## Caveat to highlight in review

The CRLF fix uses `buildBodyAst` as the normalisation seam (one `normalizeLineEndings` call at AST construction) rather than extending `normalizeBodyAst`. This was the cleanest of three options I weighed — happy to walk through the trade-off if you prefer a different placement.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,6 @@
 grammars/*.wasm filter=lfs diff=lfs merge=lfs -text
+
+# Files compared byte-for-byte against generator output on every CI
+# platform must use a stable line ending. Pin to LF so a Windows
+# checkout doesn't surprise the snapshot tests.
+docs/product/ast-fidelity-matrix.md text eol=lf
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,24 @@ jobs:
           deno-version: v2.x
       - run: deno test --allow-read --allow-write --allow-run --allow-env --allow-ffi
 
+  test-summary:
+    # Umbrella status for the `test` matrix. Branch protection on `main`
+    # requires a check literally named `Test`; the matrix expansion posts
+    # `Test (ubuntu-latest)` / `(macos-latest)` / `(windows-latest)`, so
+    # this aggregator is what satisfies the required-check rule. Adding
+    # or removing matrix legs needs no branch-protection change.
+    name: Test
+    runs-on: ubuntu-latest
+    needs: test
+    if: always()
+    steps:
+      - name: Verify matrix succeeded
+        run: |
+          if [ "${{ needs.test.result }}" != "success" ]; then
+            echo "test matrix did not succeed: ${{ needs.test.result }}"
+            exit 1
+          fi
+
   tokens:
     name: Design tokens
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,12 @@ on:
     branches: [main]
   pull_request:
 
+# Least-privilege: every job only reads the checkout. None of the CI
+# jobs comment on PRs, push code, or create releases — those workflows
+# (e.g. `release.yaml`) declare their own permissions.
+permissions:
+  contents: read
+
 jobs:
   fmt:
     name: Format
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,8 +45,12 @@ jobs:
       - run: deno check packages/markspec/main.ts packages/markspec/core/mod.ts packages/markspec/lsp/server.ts packages/markspec/mcp/server.ts
 
   test:
-    name: Test
-    runs-on: ubuntu-latest
+    name: Test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: actions/checkout@v6
         with:
@@ -75,6 +79,24 @@ jobs:
         with:
           deno-version: v2.x
       - run: bash scripts/check_ast_fidelity_matrix.sh
+
+  pwsh-installer:
+    name: PowerShell installer syntax
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Parse install.ps1
+        shell: pwsh
+        run: |
+          $errors = $null
+          [System.Management.Automation.Language.Parser]::ParseFile(
+            (Resolve-Path .\install.ps1).Path, [ref]$null, [ref]$errors
+          ) | Out-Null
+          if ($errors.Count -gt 0) {
+            $errors | ForEach-Object { Write-Error $_.Message }
+            exit 1
+          }
+          Write-Host "install.ps1 parses cleanly."
 
   commits:
     name: Conventional commits
--- a/README.md
+++ b/README.md
@@ -119,6 +119,13 @@ fn swt_brk_0001_debounce_filters_noise() {
 **Mustache variables** — `{{project.name}}` substitution from `project.yaml`,
 resolved at build time.
 
+## Install
+
+See [`docs/guide/installation.md`](docs/guide/installation.md) for the VS Code
+extension, the macOS / Linux install script, the
+[Windows PowerShell install script](docs/guide/installation.md#windows-powershell-install-script),
+manual binary downloads, and the Deno install path.
+
 ## License
 
 [MIT](LICENSE)
--- a/bootstrap
+++ b/bootstrap
@@ -37,7 +37,7 @@ detect_target() {
       esac
       ;;
     *)
-      die "unsupported OS: $os (use WSL on Windows)"
+      die "unsupported OS for bootstrap: $os (bootstrap is for contributors and needs bash; on Windows use WSL to develop, or install the released CLI with: irm https://raw.githubusercontent.com/driftsys/markspec/main/install.ps1 | iex)"
       ;;
   esac
 }
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -47,19 +47,56 @@ Add `~/.local/bin` to your `PATH` if it is not already there:
 export PATH="$HOME/.local/bin:$PATH"
 ```
 
+### Windows (PowerShell install script)
+
+Run in PowerShell 5.1 (ships with Windows 10/11) or PowerShell 7+:
+
+```powershell
+irm https://raw.githubusercontent.com/driftsys/markspec/main/install.ps1 | iex
+```
+
+The script:
+
+1. Verifies the host is x86_64. ARM Windows is not supported yet.
+2. Downloads `markspec-x86_64-pc-windows-msvc.tar.gz` and its SHA-256.
+3. Verifies the checksum with `Get-FileHash`.
+4. Extracts `markspec.exe` with `tar` (bundled on Windows 10 1803+).
+5. Places the binary in `%USERPROFILE%\.local\bin` (override with the
+   `MARKSPEC_INSTALL_DIR` environment variable).
+
+The installer does not modify your `PATH` automatically. If `markspec` is not on
+your `PATH`, the script prints both a session-scope and user-scope command. To
+make the install permanent:
+
+```powershell
+[Environment]::SetEnvironmentVariable(
+  'Path',
+  "$HOME\.local\bin;" + [Environment]::GetEnvironmentVariable('Path', 'User'),
+  'User'
+)
+```
+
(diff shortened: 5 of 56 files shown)


stasson (commented):
# Review — `feat(windows): native Windows support across CLI, LSP, and installer`

## TL;DR

**No Linux/macOS regression — approve.** CI proves it: `Test (ubuntu-latest)` passes **1648 / 0 / 1**, `Test (macos-latest)` passes **1648 / 0 / 1** — same counts, same outcomes. `Test (windows-latest)` failing is in scope of follow-up per the PR description. As a side effect this PR also brings macOS into CI for the first time, which is a net improvement for future regression detection.

## No-Linux/macOS-regression analysis

I went through each change and verified it is a no-op on POSIX inputs.

### `pathToUri` / `uriToPath` (`packages/markspec/lsp/util.ts`)

- Old: `new URL(`file://${path}`).href` / `decodeURIComponent(new URL(uri).pathname)`
- New: `toFileUrl(path).href` / `fromFileUrl(uri)`
- On POSIX, both produce identical output for absolute paths (verified by the retained test fixtures covering parens, spaces, Unicode).
- **One behavior change**: `toFileUrl` is strict — it throws on relative paths where the old code returned a malformed-but-non-throwing URL. I grepped every production caller (`definition.ts:35`, `symbols.ts:84`, `server.ts:145,159,170,630`) — all paths originate from absolute sources (`Deno.readDir` walks, parser-supplied locations, or `uriToPath` of incoming URIs). The lone offender was `symbols_test.ts` (`file: "t.md"`), which the PR updated to `"/t.md"`. No production breakage.

### `@std/path.join` substitutions

On POSIX, `join("a", "b") === "a/b"`. Pure no-op. Applied to `walkDirectory`, `walkFs`, book output paths, file-include resolution, validate-tool filtering.

### CRLF normalisation (`packages/markspec/core/util/line_endings.ts`)

The hot path guards on `if (text.indexOf("\r") < 0) return text;`, so a pure-LF file pays one `indexOf` and returns the same string. The formatter detects the source convention once (`detectLineEnding`) and skips normalisation entirely when it's `"lf"`. `applyLineEnding(text, "lf")` is identity. **A pure-LF source file goes through `format` byte-identical to before.**

Placement of the normalisation at `buildBodyAst` (rather than `normalizeBodyAst`) is the right call — it keeps the load-bearing AST-equivalence invariant intact, and the fidelity matrix surface stays at 0/58. I'd ratify the "Caveat to highlight in review" — that placement is the cleanest of the three options.

### `longestCommonDirectory` (`packages/markspec/render/mod.ts`)

Custom split-on-`/` replaced with `@std/path`'s `common()`. On POSIX, equivalent behavior. The `result === "" ? "/" : result` fallback is wrong on Windows (should be a drive root) but irrelevant on POSIX.

### `isReferencesDocument` (`packages/markspec/core/parser/mod.ts`)

Now uses `pathBasename` and `/[\\/]references[\\/]/`. On POSIX matches `/references/` — same as before.

### `relativeToRoot` (`packages/markspec/mcp/path.ts`)

Now uses `SEPARATOR` instead of literal `/`. On POSIX, `SEPARATOR === "/"` — identical.

## Issues & nits

**N1 — `${REPO_ROOT}docs/...`-style concat still in some tests** (`tests/e2e/ast_fidelity_matrix_test.ts:22`, `tests/e2e/helpers_git.ts`, e2e helpers). Author acknowledges this in the PR description ("e2e helper hardening — Deno tolerates `/` on Windows so they work, but they're not idiomatic"). Causes one of the Windows failures (`ast-fidelity-matrix: committed catalogue is not stale`). No POSIX impact.

**N2 — `Verify-Checksum` uses non-approved PowerShell verb** (`install.ps1:46`). PSScriptAnalyzer would warn. `Test-Checksum` (read-only validation) or `Confirm-Checksum` are the approved alternatives. Cosmetic; Windows-side only.

**N3 — Case sensitivity in `relativeToRoot`**. Comparison is byte-exact via `startsWith`. On Windows filesystems are case-insensitive; mismatched casing between LSP-supplied URIs and project root could miss the prefix match. Windows-side concern only.

**N4 — `onDidChangeContent` now strict on URI parsing**. `fromFileUrl` throws on malformed URIs where the old `new URL(uri).pathname` was lenient. VSCode always sends well-formed URIs, so this is arguably fail-fast goodness — but worth noting as a behavior change.

**N5 — Lone `\r` in source content gets collapsed**. `normalizeLineEndings` regex is `/\r\n?/g`, which also strips legacy bare `\r`. Theoretically a POSIX file with embedded `\r` in code blocks would lose it. In practice this is a non-issue — POSIX-authored files essentially never carry bare `\r`.

## Test coverage

- 3 new e2e tests in `format_test.ts` covering CRLF round-trip, LF preservation, and stamped-ULID-line CRLF consistency.
- 14 new unit tests in `core/util/line_endings_test.ts` — comprehensive (`detectLineEnding`, `normalizeLineEndings`, `applyLineEnding`, round-trip).
- 7 new Windows-path tests in `lsp/util_test.ts` exercised via `_pathToUriWindows` / `_uriToPathWindows` test seams — runs identically from a macOS/Linux runner. Nicely done.
- 1 new `pwsh-installer` CI job (syntax-parses `install.ps1` on every PR).

## Security (`install.ps1`)

- SHA-256 verified against the `.sha256` released alongside the tarball, before extraction.
- `LiteralPath` everywhere (no glob injection from `$HOME` or `$env:MARKSPEC_INSTALL_DIR`).
- Temp directory created with a fresh GUID and cleaned in a `finally`.
- No PATH modification — script prints commands instead. Conservative and correct.
- No code executes before checksum verification.

## Recommendation

✅ **Approve for merge** on the Linux/macOS no-regression criterion. The Windows job remains red, which is consistent with the PR framing ("Windows version may not be perfect"). Suggested follow-up sequencing:

1. Land this PR as-is — it does not regress POSIX.
2. Do **not** add `Test (windows-latest)` to required checks on `main` yet (the PR's own follow-up list says the same).
3. Address **N1** (test helper path concat) in a follow-up; that alone likely closes several Windows failures.

---

_Note: Windows job is now green following commits `4d5df8c` (full Windows support landed) and `f8a1cb1` (umbrella `Test` job to satisfy the bare-name required check). N1's concrete failure is therefore resolved; the broader "test helpers still use string-concat paths" cleanup remains a nice-to-have._

🤖 Review generated by Claude Code

PR #408, opened 2026-05-22
feat: source-view entry rendering for MarkSpec in VS Code

stasson:
## Summary

Implements [the design spec](https://github.com/driftsys/markspec/blob/feat/vscode-entry-rendering/docs/superpowers/specs/2026-05-22-vscode-entry-rendering-design.md) so MarkSpec entries read as cards in the VS Code source view.

### Visual treatment

- **Entry boundary** — 3px left bar at column 0 from title through the last trailer line (workbench colour `markspec.entry.border`, defaults to the editor's active indent-guide shade)
- **Title** — bold (decoration) on top of `class.declaration` semantic-token colour
- **Bracketed display ID** in the title line — `enum.declaration` semantic token (same colour family as the title)
- **Trailer attributes** — 60% opacity, except trace-attribute display IDs which stay bright (`enumMember` tokens) so cross-refs remain discoverable
- **Labels** — bordered rounded pills via `markspec.label.background` / `.border`; red border + hover diagnostic when the label is not in the active profile catalog
- **Modal verbs and EARS triggers** in entry body (`shall`, `should`, `may`, `must`, `will`, `When`, `While`, `If`, `Where`, `Then`) — `keyword` tokens
- **Gherkin keywords** inside `` ```feature `` / `` ```gherkin `` fenced blocks — sections (`Feature`, `Scenario`, `Background`, `Rule`, `Examples`) as `class` tokens, steps (`Given`, `When`, `Then`, `And`, `But`) as `keyword` tokens, matching GitHub Linguist and Rouge Cucumber conventions
- **`$Identifier` entity references** — `string` tokens (inline-code colour family)
- **Admonitions** — `> [!NOTE]`, `> [!TIP]`, `> [!IMPORTANT]`, `> [!WARNING]`, `> [!CAUTION]` get a 3px coloured left bar per kind and the `[!KIND]` marker renders as a coloured pill; plain `> …` blockquotes get a neutral left bar

### Architecture

- **Pure builders** in `packages/markspec/lsp/` (`entry_trailer.ts`, `semantic_tokens.ts`, `entry_ranges.ts`) with colocated unit tests; no I/O, no LSP types beyond shape-compatible local interfaces
- **LSP server** advertises `textDocument/semanticTokens/full` and a custom `markspec/entryRanges` request returning per-entry title / trailer-dim / label / blockquote / entry-boundary ranges. `validateAll()` result is cached so the custom request stays cheap on every keystroke
- **VS Code extension** (`editors/vscode/src/decorations.ts`) owns decoration types for entry border, trailer dim, title bold, valid/invalid label pills, six per-kind blockquote left bars, and five admonition marker pills; refreshed on editor change and debounced doc edit
- **Workbench colours** generated by `scripts/gen_theme.ts` from `theme/tokens.yaml`: the Paul Tol alert palette feeds the admonition kinds, the editor's indent-guide tint feeds the entry border, the diagram.blue hue feeds the label pills. `scripts/check_tokens.sh` gates against drift.

### Showcase

`docs/examples/entry-rendering.md` is expanded with EARS patterns (one entry per pattern), code blocks with entity refs, Gherkin `Scenario Outline` + `Examples` tables, parameter tables with `$variables`, `$$..$$` display math, and admonition examples for every kind.

## Test plan

- [ ] `just check` passes (1664+ tests, deno lint, dprint, four-target type-check)
- [ ] `deno test --allow-read --allow-write --allow-env packages/markspec/lsp/` — 165 unit tests pass
- [ ] `deno test --allow-read --allow-write --allow-run --allow-env tests/e2e/lsp_semantic_tokens_test.ts tests/e2e/lsp_entry_ranges_test.ts` — e2e roundtrips pass
- [ ] `cd editors/vscode && npm run compile && npm run test` — extension compiles, 19 unit tests pass
- [ ] `just tokens && git diff --exit-code` — no token drift
- [ ] Manual smoke: open `docs/examples/entry-rendering.md` in the VS Code Extension Development Host with this extension; verify entry borders, dimmed trailers, label pills, modal/EARS/Gherkin keyword highlighting, entity refs, and per-kind admonition styling render as described

## Implementation plan

[docs/superpowers/plans/2026-05-22-vscode-entry-rendering.md](https://github.com/driftsys/markspec/blob/feat/vscode-entry-rendering/docs/superpowers/plans/2026-05-22-vscode-entry-rendering.md) — executed via `superpowers:subagent-driven-development`.

## Follow-up

Epic #409 — **Body-token AST** — replaces the duplicated lexers across LSP / validator / skills with a parser-emitted token stream, and unlocks source-file (Rust `///`, Kotlin KDoc, Java javadoc, C/C++) doc-comment rendering. Currently broken because regex-level prefix detection is the wrong layer; the in-PR attempt has been reverted in favour of doing it correctly at the parser layer.

### DIFF
--- a/docs/examples/entry-rendering.md
+++ b/docs/examples/entry-rendering.md
@@ -15,9 +15,22 @@ attribute is written in source. Cross-reference relation names (`Satisfies`,
   falls below the configurable threshold and the driver has not applied the
   brake pedal.
 
+  ```feature
+  Scenario: Noise spike shorter than debounce window
+    Given a debounce window of 10ms
+    And a stable pressure reading of 500
+    When a spike of 999 occurs for 5ms
+    Then the output shall remain 500
+
+  Scenario: Sustained change longer than debounce window
+    Given a debounce window of 10ms
+    And a stable pressure reading of 500
+    When the reading changes to 600 for 15ms
+    Then the output shall change to 600
+  ```
+
       Id: 01HGW3A2BCD5VWXYZABCDEFGHJ
-      Labels: ASIL-B
-      Labels: safety-critical
+      Labels: ASIL-B, Labels: safety-critical
 
 - [SYS_AEB_0012] Object threat assessment from radar tracks
 
@@ -28,16 +41,16 @@ attribute is written in source. Cross-reference relation names (`Satisfies`,
       Satisfies: STK_AEB_0001
       Labels: ASIL-B
 
+
 - [SWE_BRK_0107] Median filter implementation
 
   The braking ECU shall apply a 5-sample median filter to the raw brake pressure
-  sensor input before processing.
+  sensor input before processing.  $aNiceVariable
+
 
       Id: 01HGW2Q8MNP3RSTVWXYZABCDEF
       Satisfies: SYS_AEB_0012
-      Labels: ASIL-B
-      Labels: real-time
-      Labels: performance
+      Labels: ASIL-B, real-time, performance
 
 Prose between entries renders normally — no left border, no type coloring. The
 visual separation between entries and prose is provided by the admonition border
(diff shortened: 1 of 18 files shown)


stasson:
## Code Review

### Overview

The PR cleanly factors source-view rendering into pure builders (`entry_trailer`, `entry_ranges`, `semantic_tokens`) wired into the LSP via a standard semantic-tokens provider plus a `markspec/entryRanges` custom request. The VS Code side is a thin `DecorationManager` over three `TextEditorDecorationType`s with theming sourced from `theme/tokens.yaml`. The split between portable LSP semantic tokens (themes do the work) and a VS-Code-specific decorations channel for pills is a good architectural choice.

### Strengths

- **Pure builders + colocated unit tests.** Trailer scanner, semantic-token builder, and entry-ranges builder are all pure functions with sibling `*_test.ts` files, matching the project convention. No I/O inside the builders, no LSP types leaking in beyond shape-compatible local interfaces.
- **Grammar consistency.** The display-ID regex in `entry_trailer.ts` mirrors `hover.ts` / `rename.ts` (alphanumeric start, `[A-Za-z0-9._/-]`, ≥3 chars). Regression tests in `semantic_tokens_test.ts` lock that in.
- **Tokens SSOT preserved.** Extending `gen_theme.ts` to write `contributes.colors` and gating it via `check_tokens.sh` keeps the Paul Tol palette as the single source — same convention as the Typst/CSS outputs.
- **LSP wire encoder is correct.** `encodeSemanticTokens` uses `deltaStart = startChar - prevChar` (start-to-start) and updates `prevChar = t.startChar` — matches the LSP spec.
- **Sensible defaults.** Both new handlers guard with `isMarkspecFile` and return empty payloads when the document isn't tracked; `DecorationManager.refresh` bails silently on `sendRequest` failure.

### Issues

#### Important (worth addressing in follow-up)

- **`server.ts` (entryRanges handler) — full-workspace revalidation on every request.** The handler calls `index.validateAll()`, which runs the entire validator across all indexed entries every time. The 100 ms client-side debounce limits the burst, but this duplicates the background validation already publishing diagnostics every 1000 ms. Suggested fix: cache the latest `validateAll()` result alongside the debounced publish (e.g. `let lastDiagnostics: readonly Diagnostic[] = []`), and let the handler read from that cache. Alternative: filter diagnostics to the current file before passing into `buildEntryRanges` and only walk per-file diagnostics in `findDiagnosticAt`.

- **`entry_ranges.ts` (`computeDimRanges`) — `Id:` value is excluded from dim ranges, leaving the ULID visually bright.** The trailer scanner returns the ULID inside `idRanges` (it satisfies the display-ID grammar), and `computeDimRanges` excludes every `idRanges` entry regardless of attribute key. The design says only display IDs in *trace attributes* stay bright; `Id:` is the identity slot. Compare with `semantic_tokens.ts`, which correctly special-cases `Id:` to emit a single `string` token. The two render paths now disagree on the same trailer line. Fix: in `computeDimRanges`, treat `tl.key === "Id"` (and `tl.key === "Labels"`, since pills handle those separately) as "no ID exclusions" — the whole value is dim for `Id`. Add a unit test asserting the `Id:` line has one contiguous dim range.

- **`entry_trailer.ts` — `valueLength` includes trailing whitespace.** `valueLength = line.length - valueStart` extends the value past actual content if the line has trailing spaces. Visually invisible for `string` tokens, but the dim region drawn past EOL may produce a faint glyph on some themes. Fix: trim trailing whitespace from `valueLength` (or trim at the call site).

- **Free-form attribute values matching the display-ID grammar are highlighted as cross-references.** `Note: foobar` produces an `enumMember` token treating `foobar` as a trace target. The only signal is the value's shape, not the key's semantics. This is consistent with how `hover.ts` / `rename.ts` already behave, so not a regression — but the new render layer makes it visually obvious. Worth a TODO or follow-up issue to consult the profile's attribute declarations.

#### Minor

- **`scripts/gen_theme.ts` — round-tripping `editors/vscode/package.json` reformats unrelated keys.** `JSON.stringify(pkg, null, 2)` expanded `"items": { "type": "string" }` and `"default": ["lsp"]` to multi-line form. The drift gate will fail any manual single-line edit. The diff already shows the side-effect. Consider documenting it in `gen_theme.ts` so future editors don't compress the JSON back to one-liners.

- **`extension.ts` — listeners registered after `client.start()`.** vscode-languageclient buffers notifications so this works in practice, but registering before `start()` would be safer. Same applies to the initial `decorations.refresh(window.activeTextEditor)` call — the silent `catch` in `refresh()` covers the timing window, but `await client.onReady?.()` (or `client.start().then(...)`) would make the order explicit.

- **`entry_ranges.ts` (`findDiagnosticAt`) is O(diagnostics × labels).** For a workspace with many cross-file diagnostics, this nested scan is wasteful. Bucket diagnostics by file/line once at the top of `buildEntryRanges`. Not urgent at current scale.

- **`semantic_tokens.ts` (`splitWithOffsets`) doesn't handle CSV-with-citations.** Top-level commas inside `[…]` would split wrong for citation locators. Labels don't carry citations today; `code_actions.ts:buildA011Fix` already has the depth-aware split if the convention spreads.

- **`extension.ts` — `debounce` is duplicated** with `lsp/util.ts`. The Node-side extension cannot import the Deno-only module, so this is justified — a one-line comment noting the deliberate duplication would prevent a future "DRY this" cleanup that breaks the type boundary.

- **`semantic_tokens.ts` — `enum` and `enumMember` used for both title display ID and labels.** Same token type for very different concepts means a theme cannot color them independently. Acceptable trade-off for legend size; flag in a follow-up.

### Test coverage assessment

**Covered well:**
- Trailer scanner: 6 tests (Id row, multi-ID Satisfies, multi-space-after-colon, multi-entry boundary, broad-grammar, no-trailer empty).
- Semantic-tokens builder: 11 tests including legend, title, Id special-case, trace-attr, label validity, empty-profile, multi-entry scoping, plain-string fallback, and three regressions for broader title grammar.
- Entry-ranges builder: 4 tests (title range, dim-range ID exclusion, label validity, diagnostic attachment).
- E2E: both endpoints exercised through the real LSP wire.

**Gaps worth filling:**
- Multi-entry input for `entry_ranges` (`endLineExclusive` slicing).
- `Id:` line dim behavior (would have caught the inconsistency flagged above).
- Empty document / no entries → safe empty response.
- Malformed entry where the title line does not match `TITLE_LINE_RE` (skip path).
- Labels with all-whitespace segment / trailing comma.

### Final assessment

🟡 **Approved with non-blocking suggestions.**

No critical bugs and no security concerns. The most impactful items are the `validateAll()` cost per keystroke and the `Id:`-stays-bright inconsistency between the two render paths — both fixable in small follow-ups and not regressions of existing behavior.

---

Generated by `/review` on PR #408.

PR #411, opened 2026-05-23
feat(lsp): stamp ULIDs in scaffold completions; add trailer-key completions

stasson:
## Summary

Stage 1 of the LSP completion improvements design.

- **Block-scaffold completion now stamps a real ULID** into the `Id:` line at insert time. No more post-insert `markspec format` step.
- **`completionItem/resolve` re-queries** `WorkspaceIndex.getNextDisplayIdNumber` and regenerates the ULID at resolve time, so two rapid back-to-back inserts get consecutive display IDs (the index re-parses on a 50 ms debounce; without resolve, the second list can show a stale number).
- **New trailer-key completions**: typing `Sa` (etc.) on an indented blank line inside an entry trailer offers `Satisfies:`, `Verified-by:`, and the other 11 trailer attribute keys. Wired as a fourth trigger in `onCompletion`.

Design spec: `docs/superpowers/specs/2026-05-23-completion-ulid-and-inline-ai-design.md` — Stage 1 only. Stage 2 (inline AI via `vscode.lm`) ships in a follow-up PR after this one merges.

## Commits

Six commits, all conventional with `(lsp)` scope:

- `c64f343` feat(lsp): stamp real ULID in block-scaffold completions
- `13bb46b` refactor(lsp): document ulidProvider param and tighten tests
- `5a82235` feat(lsp): resolve scaffold completions with monotonic next-id
- `f12e547` refactor(lsp): type the scaffold resolve data payload
- `83db361` feat(lsp): suggest trailer attribute keys in entry trailer region
- `cb8d3a9` refactor(lsp): clarify trailer-key trigger docs and tighten tests

## Test plan

Automated checks all green locally:

- [x] `just build` (lint + test + typecheck + compile) — 1675 tests pass
- [x] `deno fmt --check`
- [x] `dprint check`

Manual VSCode smoke tests for the reviewer (could not run in the authoring session):

- [ ] Open a MarkSpec markdown file in VSCode with the freshly-compiled `dist/markspec` binary. Type `- [` on a new line. Accept the suggested scaffold. The inserted `Id:` line shows a 26-char Crockford base32 ULID, not the literal placeholder.
- [ ] Run `markspec validate` on the file after the insert. It passes without first running `markspec format`.
- [ ] Insert two scaffold completions back-to-back. The second display ID is exactly one higher than the first.
- [ ] Inside an entry trailer region (indented 6 spaces, blank line), type `Sa`. The completion list offers `Satisfies:`. Accept it — the cursor lands after the colon-space.

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/packages/markspec/lsp/completions.ts
+++ b/packages/markspec/lsp/completions.ts
@@ -3,10 +3,11 @@
  *
  * Completion providers for MarkSpec entry blocks and ID references.
  *
- * Three triggers:
+ * Four triggers:
  * 1. Block scaffold — `- [` at line start → full entry block snippet
- * 2. ID reference — trace attribute keyword (e.g., `Satisfies:`) → display ID list
- * 3. Type: attribute value — `Type:` keyword → core and profile type list
+ * 2. Trailer attribute key — indented blank or partial capitalized key → key list
+ * 3. ID reference — trace attribute keyword (e.g., `Satisfies:`) → display ID list
+ * 4. Type: attribute value — `Type:` keyword → core and profile type list
  *
  * Block-scaffold items receive a fresh ULID from a `ulidProvider` callback so
  * insertions don't need a follow-up `markspec format` pass.
@@ -27,6 +28,42 @@ const TRACE_ATTR_RE =
 
 /** Pattern matching the `Type:` attribute keyword at line start. */
 const TYPE_ATTR_RE = /^\s*Type\s*:/;
+
+/**
+ * Trailer attribute keys offered as completions inside an entry's
+ * trailer region. Includes the eleven trace-link keys (whose pattern
+ * also appears in `TRACE_ATTR_RE` and in `TRACE_KEYWORDS_RE` in
+ * `context.ts`), plus the cardinal `Labels:` and `Type:` keys. When
+ * the trace-keyword set changes here, also update those two regexes.
+ */
+export const TRAILER_KEYS: readonly string[] = [
+  "Satisfies",
+  "Derived-from",
+  "Verified-by",
+  "References",
+  "Tests",
+  "Depends-on",
+  "Part-of",
+  "Allocated-to",
+  "Realizes",
+  "Generated-from",
+  "Supersedes",
+  "Labels",
+  "Type",
+] as const;
+
+/** Trailer-region context: indent ≥4 spaces, optional partial capitalized key. */
+const TRAILER_KEY_CONTEXT_RE = /^\s{4,}([A-Z][A-Za-z-]*)?$/;
+
+/**
+ * Check whether the text before the cursor is in an entry's trailer
+ * region and ready to receive an attribute key. Matches a line that
+ * starts with at least 4 spaces of indent and contains nothing or a
+ * partial capitalized key.
+ */
+export function isTrailerKeyContext(textBefore: string): boolean {
+  return TRAILER_KEY_CONTEXT_RE.test(textBefore);
+}
 
 /**
  * Check if the text before cursor triggers a block scaffold completion.
@@ -77,6 +114,8 @@ export interface CompletionItemData {
 /** CompletionItemKind.Snippet = 15, CompletionItemKind.Reference = 18 */
 const KIND_SNIPPET = 15;
 const KIND_REFERENCE = 18;
+/** LSP `CompletionItemKind.Property` numeric value (10). */
+const KIND_PROPERTY = 10;
 
 /** Entry type info for block scaffold completion. */
 export interface EntryTypeInfo {
@@ -140,6 +179,21 @@ export function buildTypeAttributeItems(
     });
   }
   return items;
+}
+
+/**
+ * Build completion items for the trailer-key trigger. One item per
+ * entry in {@linkcode TRAILER_KEYS}, each inserting `<Key>: ` with
+ * the cursor placed after the colon.
+ */
+export function buildTrailerKeyItems(): CompletionItemData[] {
+  return TRAILER_KEYS.map((key) => ({
+    label: key,
+    detail: "trailer attribute",
+    insertText: `${key}: `,
+    isSnippet: false,
+    kind: KIND_PROPERTY,
+  }));
 }
 
 /** Inputs to {@linkcode renderScaffoldSnippet}. */
--- a/packages/markspec/lsp/completions.ts
+++ b/packages/markspec/lsp/completions.ts
@@ -52,7 +52,7 @@ export const TRAILER_KEYS: readonly string[] = [
   "Type",
 ] as const;
 
-/** Trailer-region context: indent ≥4 spaces, optional partial capitalized key. */
+/** Trailer-region context: indent ≥4 whitespace chars (matches the parser's lenient leading-whitespace acceptance; formatter canonicalises to 6 spaces), optional partial capitalized key. */
 const TRAILER_KEY_CONTEXT_RE = /^\s{4,}([A-Z][A-Za-z-]*)?$/;
 
 /**
@@ -186,6 +186,7 @@ export function buildTypeAttributeItems(
  * entry in {@linkcode TRAILER_KEYS}, each inserting `<Key>: ` with
  * the cursor placed after the colon.
  */
+// Kept as a zero-arg function (rather than a precomputed constant) for naming symmetry with the other `build*Items` helpers and so future profile-aware filtering can be added without changing the API.
 export function buildTrailerKeyItems(): CompletionItemData[] {
   return TRAILER_KEYS.map((key) => ({
     label: key,
--- a/packages/markspec/lsp/completions.ts
+++ b/packages/markspec/lsp/completions.ts
@@ -147,6 +147,7 @@ export function buildTypeAttributeItems(
 
 export function buildBlockScaffoldItems(
   types: readonly EntryTypeInfo[],
+  ulidProvider: () => string,
 ): CompletionItemData[] {
   if (types.length === 0) {
     return [
@@ -162,11 +163,12 @@ export function buildBlockScaffoldItems(
 
   return types.map((type) => {
     const displayId = `${type.prefix}${padNumber(type.nextNumber)}`;
+    const ulid = ulidProvider();
     return {
       label: `New ${type.name} (${displayId})`,
       detail: type.name,
       insertText:
-        `${displayId}] \${1:Title}\n\n  \${2:Body.}\n\n      Id: \\$\{ULID}\n      \${3:Satisfies: }`,
+        `${displayId}] \${1:Title}\n\n  \${2:Body.}\n\n      Id: ${ulid}\n      \${3:Satisfies: }`,
       isSnippet: true,
       kind: KIND_SNIPPET,
     };
--- a/packages/markspec/lsp/completions.ts
+++ b/packages/markspec/lsp/completions.ts
@@ -150,11 +150,28 @@ export interface ScaffoldSnippetInput {
   readonly ulid: string;
 }
 
+/** Discriminator value for scaffold completions' resolve-time `data` payload. */
+export const SCAFFOLD_COMPLETION_KIND = "scaffold";
+
+/**
+ * `data` payload attached to scaffold completion items so the
+ * `completionItem/resolve` handler can re-query the workspace index
+ * and regenerate the snippet with the freshest display ID + ULID.
+ */
+export interface ScaffoldCompletionData {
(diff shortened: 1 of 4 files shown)


stasson:
## Code review

### Overview

Stage 1 LSP completion improvements: stamps real ULIDs into block-scaffold completions, adds a `completionItem/resolve` handler that re-queries the workspace index for fresh display IDs + ULIDs at resolve time, and introduces a new trailer-attribute-key trigger. Diff is +349/-25 across 4 files (`lsp/completions.ts`, `lsp/server.ts`, `lsp/completions_test.ts`, `lsp/workspace_test.ts`). Six commits, all Conventional with `(lsp)` scope. No protocol changes, no new dependencies, no `core/` touched — the boundary is preserved.

### Code quality

**Strengths**
- Clean separation: pure builders + helpers live in `completions.ts`; LSP protocol wiring stays in `server.ts`. The new `renderScaffoldSnippet` helper is the single source of truth for both build-time and resolve-time rendering, eliminating drift.
- The `data: ScaffoldCompletionData` payload + `SCAFFOLD_COMPLETION_KIND` discriminator is typed end-to-end. `satisfies ScaffoldCompletionData` at the construction site catches widening; the cast at the resolve site narrows safely.
- TDD followed throughout — every commit has failing-test-first evidence. Test coverage at the diff level: 13 new test cases (5 for ULID, 1 for `renderScaffoldSnippet`, 1 for `WorkspaceIndex` next-id, 8 for trailer keys, plus 2 existing tests updated for the new signature).
- Conventional Commits applied with consistent `feat(lsp)` / `refactor(lsp)` split — three feat commits each followed by a refactor cleanup commit. PR is easy to read commit-by-commit.

**Issues**

*Minor — `isTrailerKeyContext` JSDoc still says "4 spaces" while the regex constant doc says "4 whitespace chars".* In `packages/markspec/lsp/completions.ts`, the constant's doc comment was updated to "indent ≥4 whitespace chars" but the JSDoc on `isTrailerKeyContext` immediately below still reads "starts with at least 4 spaces of indent". The two should match — recommend tightening the function JSDoc to "≥4 leading whitespace chars" for consistency.

*Minor — `// Kept as a zero-arg function ...` is a long inline comment that would read better as a `@remarks` JSDoc tag.* In `buildTrailerKeyItems`, the explanation about not being a precomputed constant lives as a single-line `//` comment between the JSDoc block and the function signature. Moving it into the JSDoc body would keep all of the function's documentation in one place.

*Minor — `as unknown as` would be safer than `as ScaffoldCompletionData | undefined`.* In `onCompletionResolve`, `item.data` is typed `any` by the LSP library; the direct cast assumes the shape rather than validating it. The narrow `kind` check that follows mitigates this, but a defensive `as unknown` intermediate cast would make the type-system contract explicit. Low priority — the check is sufficient at runtime.

### Project conventions

- ✅ `feat(lsp)` / `refactor(lsp)` scoped Conventional Commits
- ✅ Pure module pattern preserved (all of `completions.ts` is exported testable functions)
- ✅ `core/mod.ts` boundary respected — no changes outside `lsp/`
- ✅ `@std/ulid` imported from the JSR map (same source the formatter already uses)
- ✅ `data` payload kinds are named constants, not magic strings

### Performance

- `onCompletionResolve` calls `getNextDisplayIdNumber` (O(N) over indexed entries) once per accepted item. At workspace sizes typical of MarkSpec projects this is negligible. If the workspace grows to thousands of entries the linear scan could become measurable; not a concern for now.
- `buildTrailerKeyItems` allocates 13 new objects on every keystroke that hits the trigger. The current call frequency is bounded by the user's typing speed inside trailers, so the allocation cost is well under a microsecond. Reviewer flagged this as a possible memoization opportunity; the implementer's response (keep the function shape for future profile-aware filtering) is reasonable.

### Test coverage

- All three new behaviors have unit tests in `completions_test.ts`.
- The resolve handler's index dependency is exercised via a `WorkspaceIndex` integration test in `workspace_test.ts` — the right placement (the test exercises the index contract, not the LSP wire).
- The four manual VSCode smoke tests in the PR description require a live editor and were not run in the authoring session. **Reviewer should run them before merge.**

### Security

- No new external inputs accepted.
- `ulid()` from `@std/ulid` is cryptographically random — no injection surface in the snippet construction.
- The `item.data` cast in `onCompletionResolve` is from an LSP-controlled object; even if a hostile client sent a tampered `data` payload, the worst case is a no-op (the kind discriminator filters it).

### Verdict

**Approve pending the manual smoke tests in the PR description.** The minor doc inconsistencies above can be folded into a follow-up commit on this branch or deferred — none are blocking.

stasson:
## Formal review

This formal review consolidates findings from a dedicated security pass and a comprehensive code-quality pass.

---

### Summary

Stage 1 of the LSP completion improvements design. Six commits across four files (+349/-25). Three behaviors land: real ULID stamping at completion-build time, a `completionItem/resolve` handler for monotonic next-display-id, and a new trailer-attribute-key trigger.

Architecture is clean — `completions.ts` stays protocol-free, `server.ts` handles all LSP wiring. The `renderScaffoldSnippet` helper is the right level of extraction (genuinely shared between build-time and resolve-time paths). The `ScaffoldCompletionData` discriminator pattern with `typeof SCAFFOLD_COMPLETION_KIND` is type-checked end-to-end. Test coverage targets behavior, not just surface.

### Verdict

**APPROVE WITH MINOR CHANGES** — one protocol correctness bug and two security hardening items should be addressed before merge; everything else is small.

---

### Strengths

- **Module-boundary discipline preserved.** Pure builders in `completions.ts`, protocol wiring in `server.ts`. `core/mod.ts` untouched. No new dependencies (`@std/ulid` was already a workspace dep used by the formatter).
- **Discriminator + `satisfies` pattern.** `data: { kind: SCAFFOLD_COMPLETION_KIND, ... } satisfies ScaffoldCompletionData` at `server.ts:485` plus the typed cast at `server.ts:529` give an end-to-end type-checked handshake. Future trigger kinds can follow the same pattern.
- **Trigger ordering is correct and disjoint.** `isTrailerKeyContext` requires no colon; `isTraceAttributeTrigger` requires a colon to be present. Trigger 2 vs Trigger 3 don't race.
- **Tests pin behavior, not surface.** Three `buildBlockScaffoldItems` tests cover ULID-baked, provider-called-per-item, and zero-types fallback. `renderScaffoldSnippet` test is shape-aware (`startsWith` + `Id: <ulid>` substring). The `WorkspaceIndex` integration test exercises the right contract.
- **Commit hygiene.** Three feat/refactor pairs follow the same shape per feature. Conventional Commits with `(lsp)` scope throughout.

---

### Findings

#### Critical

**C1. `onCompletionResolve` does not set `insertTextFormat: Snippet` — resolved items may insert snippet syntax literally on clients that strip the item before resolve.**

`server.ts:537-541`:

```typescript
return {
  ...item,
  label: rendered.label,
  insertText: rendered.insertText,
};
```

The spread `...item` preserves `insertTextFormat` only if the LSP client echoes the full item back in the resolve request. The LSP specification (§3.16.2) doesn't guarantee this — clients may send a stripped item containing only `data` and identity fields. If `insertTextFormat` is absent in the resolved response, the editor defaults to `PlainText`, and the placeholder text `${1:Title}` lands literally in the document.

The original `onCompletion` path sets `insertTextFormat: InsertTextFormat.Snippet`. The resolve handler must do the same:

```typescript
return {
  ...item,
  label: rendered.label,
  insertText: rendered.insertText,
  insertTextFormat: InsertTextFormat.Snippet,
};
```

VSCode currently echoes the full item so this doesn't manifest there, but Neovim/coc.nvim, Helix, and Zed implement resolve and may strip the item.

#### Important

**I1. `onCompletionResolve`: `data.prefix` and `data.typeName` are not type/length validated after the discriminator check.**

`server.ts:527-543`. The gate at `data?.kind !== SCAFFOLD_COMPLETION_KIND` proves `kind === "scaffold"` but makes no assertion about the runtime types or sizes of `prefix` and `typeName`. Both fields are then:

- Passed to `getNextDisplayIdNumber(prefix)` which calls `String.prototype.startsWith(prefix)` on every indexed display ID. An unusually long `prefix` blocks the event loop in proportion to `|prefix| × N(entries)`.
- Interpolated into `insertText` and `label` template literals. On a well-behaved client this is fine; a hostile or buggy client could send tampered fields that produce surprising editor behavior.

Recommended fix at the top of the resolve handler:

```typescript
if (
  typeof data.prefix !== "string" || data.prefix.length > 64 ||
  typeof data.typeName !== "string" || data.typeName.length > 128
) {
  return item;
}
```

The 64/128 caps are well above any realistic profile prefix or type name.

**I2. `renderScaffoldSnippet`: `prefix` and `typeName` interpolated into LSP snippet syntax without escaping `$` or `\`.**

`completions.ts:231-239`. If a profile-declared prefix or type name contains `${` or `\`, that text lands unescaped inside `insertText` where the editor's snippet engine parses it. The ULID field is safe (`@std/ulid` emits only uppercase alphanumerics + digits). Profile values come from the user's own `project.yaml` so this is low-severity in practice — but a repository with a crafted profile could nudge the cursor to an unexpected snippet position.

Recommended hardening:

```typescript
function escapeSnippet(s: string): string {
  return s.replace(/\\/g, "\\\\").replace(/\$/g, "\\$");
}
// then use escapeSnippet(input.prefix) / escapeSnippet(input.typeName) in interpolations
```

**I3. `TRAILER_KEYS` omits `Id` without documenting why — future maintainers may add it by mistake.**

`completions.ts:32-53`. `Id:` is intentionally excluded because the formatter stamps it automatically, and hand-written ULIDs fail validation (per `~/.claude/rules/markspec-core-rules.md`). This invariant should be stated in the JSDoc above `TRAILER_KEYS`:

> `Id:` is intentionally excluded — the formatter stamps it automatically, and hand-written ULIDs fail validation.

**I4. The inline `// Kept as a zero-arg function ...` comment on `buildTrailerKeyItems` sits between the JSDoc and the function signature.**

`completions.ts:184-191`. Unusual position — other modules either fold this kind of note into the JSDoc body (e.g., as a closing paragraph) or place it as the first comment inside the function body. Move it to keep the JSDoc/signature pairing clean.

#### Minor

**M1. `TRAILER_KEY_CONTEXT_RE` doc comment is a single very long line.** `completions.ts:55`. The line exceeds 120 chars where the rest of the module uses multi-line JSDoc blocks. `dprint` doesn't reflow comments so this passes CI but is stylistically inconsistent.

**M2. No `isTrailerKeyContext` test for tab indentation.** `completions_test.ts:223-227` covers 4-space and 6-space indents, but not `\t`. The regex `\s{4,}` accepts tabs; the formatter canonicalises to spaces. Either add a test documenting the tab acceptance (or, if the intent is space-only, tighten the regex to `\x20{4,}` and add a rejection test).

**M3. `renderScaffoldSnippet` test doesn't pin tab-stop ordering.** `completions_test.ts:186-201`. Asserts the label, the bracket opener, and the `Id:` line — but not that tab stops appear as `${1:Title}` → `${2:Body.}` → `${3:Satisfies: }`. A regression that swapped two tab stops would slip through. Add `assertStringIncludes(snippet.insertText, "${1:Title}")` etc.

**M4. `SCAFFOLD_COMPLETION_KIND` and `ScaffoldCompletionData` are exported but have no external consumers.** `completions.ts:208-220`. They're public for `server.ts`'s benefit, but the export also invites unrelated callers. A brief `@internal` JSDoc tag or a module-doc note ("exported for `server.ts` only") would set the right expectation.

---

### Observations (non-findings)

- **Resolve-time re-query is best-effort, not guaranteed.** Clients aren't required to call `completionItem/resolve` before inserting. A client that accepts directly from the initial list uses the stale (build-time) values. The PR is still a meaningful improvement for clients that do resolve (VSCode, Zed); the problem isn't fully closed until a future stage (e.g., post-insert reformat) lands. Stage 1's value stands on its own.
- **`isTrailerKeyContext` will false-trigger on deeply-indented body content** (e.g., a nested list item indented 4+ spaces). The body region is conventionally 2-space indented and the trailer 4+, so the line-prefix heuristic works in practice — but the false positive is easy to surface and dismiss. Worth a note in the function JSDoc.
- **`@std/ulid` is locked in `deno.lock` with an integrity hash and was already a transitive dep on `main`.** No new trust boundary.
- **No new network calls, file reads, or shell-outs.** Confirmed by full-diff inspection.

---

**Recommended action:** Address C1, I1, and I2 in one follow-up commit on this branch before merging. I3/I4 can be folded into the same commit (doc-only changes). Minor items can ship as-is or be batched with the next LSP touch.

PR #425, opened 2026-05-24
docs: remove aspice-slice placeholder from examples guide

### LINKED ISSUE #419: docs(examples): aspice-slice corpus — schedule, or drop the promise from the guide?
[docs/guide/examples.md](../blob/main/docs/guide/examples.md) is a
"Coming soon" stub that promises:

> The gallery will be grounded in the end-to-end test corpus at
> `tests/fixtures/corpora/aspice-slice/`, which does not yet exist in the
> repository.

The corpus doesn't exist; the guide chapter is empty as a result. The
only real example today is
[docs/examples/entry-rendering.md](../blob/main/docs/examples/entry-rendering.md),
which covers entry block rendering but not multi-file projects, listing
directives, compile-output, or doc-comment entries in source files.

## Question

Is the aspice-slice corpus a near-term priority worth scheduling, or
should we drop the promise from `docs/guide/examples.md` and ship a
different example (e.g., a thin slice that uses what already exists)?

The highest-value gap that isn't covered anywhere else is a source-file
doc-comment example (e.g., a Rust crate or Kotlin module with MarkSpec
entries in `///`/`/** */` comments) — that's the workflow most unique to
MarkSpec.

## Acceptance

Either:

- The aspice-slice corpus is scheduled (target version + owner), and the
  guide page stays as-is until it lands, OR
- The corpus promise is removed from `docs/guide/examples.md` and a
  replacement example scope is agreed.

Spun out from #184 bucket 4.

narutamaaurum:
## Summary
- replace the placeholder examples page with links to examples that exist today
- remove the promise that `tests/fixtures/corpora/aspice-slice/` already exists
- keep a short outline for what a future end-to-end corpus chapter could cover

## Validation
- git diff --check

Closes #419

### DIFF
--- a/docs/guide/examples.md
+++ b/docs/guide/examples.md
@@ -1,25 +1,28 @@
 # Examples gallery
 
-> **Coming soon.**
->
-> This chapter will tour the in-repo example project — a thin slice of an
-> automotive emergency-braking system — so you can clone it and modify it on day
-> one.
+MarkSpec does not yet ship the larger in-repo `aspice-slice` corpus that older
+planning notes referenced, so this chapter focuses on the examples you can use
+right now.
 
-The gallery will be grounded in the end-to-end test corpus at
-`tests/fixtures/corpora/aspice-slice/`, which does not yet exist in the
-repository. Once the corpus ships, this chapter will cover:
+## Available examples today
+
+- **[Quickstart](quickstart.md)** — build a small self-contained project from
+  scratch and validate it locally.
+- **[`docs/examples/entry-rendering.md`](../examples/entry-rendering.md)** — see
+  every entry block rendered with the current typography and pill treatments.
+
+## What to expect from a future corpus example
+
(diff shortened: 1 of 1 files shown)


stasson:
@narutamaaurum  thanks for your contribution, I'm in the middle of the dev phase so I wont take it now. once I have reached something stable I'll consider it.

stasson:
Parking this pending a deliberate minimal-vs-richer decision, tracked in #678.

Context: this PR's `Closes #419` reference is dead — #419 was closed `NOT_PLANNED` on 2026-05-25 with the decision to *drop the aspice-slice promise from the guide*, which is exactly what this PR implements. It's still mergeable and the premise still holds (the corpus dir still doesn't exist; the placeholder is still live on `main`), so if the minimal option is chosen it can be refreshed (fix the dead issue link) and merged as-is. See #678.

PR #461, opened 2026-05-25
feat(parser): add C# grammar support

stasson:
## Summary

- Adds C# (`.cs`) to the source-parser dispatch via `tree-sitter-c-sharp@0.23.1`.
- Wires `csharp` through `SupportedLanguage`, `EXT_TO_GRAMMAR`, `languageIdForExtension`, `SOURCE_EXTENSIONS`, and `LANGUAGE_SPECS`.
- New `csharpItemName` extractor handles three exceptions to the simple `name`-field path: `operator_declaration` / `indexer_declaration` return undefined; `field_declaration` / `event_field_declaration` walk into the inner `variable_declarator` (first declarator wins for multi-name decls). `destructor_declaration` is intentionally NOT special-cased — the grammar exposes the class identifier as the destructor's `name` field, so `~Outer()` yields `"Outer"`.
- Fetch script gains an optional `urlFile` on `NpmGrammar` to handle the upstream package's underscore naming (`tree-sitter-c_sharp.wasm`) while keeping the on-disk hyphen-naming convention.
- 10 new unit tests in `source_test.ts` characterize: basic XML doc-line extraction, Javadoc block fallback, per-item-type happy path (14 entries covering class/struct/interface/record/record-struct/enum/delegate/method/property/constructor/local-function/nested-local-function/file-scoped-namespace), destructor returning class name, operator/indexer returning undefined, field/event-field/multi-name name extraction, and the regular `//` negative case.
- New `tests/fixtures/in-code-csharp.cs` mirrors the existing in-code Java/C++/Rust fixtures.

## Version pin rationale

`tree-sitter-c-sharp@0.23.5` (latest) is compiled against tree-sitter language ABI v15; `web-tree-sitter@^0.24` caps at v14. Pinning to **0.23.1** (newest ABI-v14 grammar). Upgrading web-tree-sitter is out of scope for this PR.

## Test plan

- [x] `just build` (lint + test + typecheck + compile)
- [x] `deno fmt --check`
- [x] `dprint check`
- [x] Full `source_test.ts` suite: 44 passed, 0 failed
- [x] Manual: `markspec validate tests/fixtures/in-code-csharp.cs` exits 0 (with expected profile-coverage warnings, same as Java/C++)

## Parallel-safe with JS-family PR

A sibling sub-project (`docs/superpowers/2026-05-25-js-family-grammars-handoff.md`) is adding TypeScript/TSX/JavaScript in parallel. Both PRs touch the same eight files (`SupportedLanguage`, `EXT_TO_GRAMMAR`, `SOURCE_EXTENSIONS`, `LANGUAGE_SPECS`, `fetch_grammars.ts`, `grammars/README.md`, `AGENTS.md`, `grammars.lock`), all additive — conflicts will be trivial. Merge order coordinated at PR time.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -323,7 +323,7 @@ profile (ASPICE, ISO 26262) would declare additional types
 - **Language:** TypeScript (strict mode)
 - **Markdown parsing:** unified / remark / mdast ecosystem
 - **Source file parsing:** tree-sitter (for doc comment extraction from Rust,
-  Kotlin, Java, C, C++, TypeScript, TSX, JavaScript)
+  Kotlin, Java, C, C++, TypeScript, TSX, JavaScript, C#)
 - **PDF rendering:** Typst via typst.ts WASM embedding
 - **Book rendering:** custom static site generator (`book/` module)
 - **Presentations:** Touying (Typst presentation framework)
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -660,8 +660,8 @@ defined in your profile.
 in the project.
 
 **Source file context guard** — in source files (Rust, Kotlin, Java, C, C++,
-TypeScript, TSX, JavaScript), completions only activate near entry markers or
-trace keywords. The server won't interfere with your language's native LSP
+TypeScript, TSX, JavaScript, C#), completions only activate near entry markers
+or trace keywords. The server won't interfere with your language's native LSP
 (rust-analyzer, kotlin-lsp, etc.).
 
 ### VS Code
--- a/editors/vscode/src/extension.ts
+++ b/editors/vscode/src/extension.ts
@@ -127,6 +127,7 @@ export function activate(context: ExtensionContext): void {
       { scheme: "file", language: "typescriptreact" },
       { scheme: "file", language: "javascript" },
       { scheme: "file", language: "javascriptreact" },
+      { scheme: "file", language: "csharp" },
     ],
     synchronize: {
       fileEvents: workspace.createFileSystemWatcher("**/*.md"),
--- a/grammars/README.md
+++ b/grammars/README.md
@@ -21,6 +21,7 @@ Supported grammars:
 | `tree-sitter-typescript.wasm` | TypeScript             |
 | `tree-sitter-tsx.wasm`        | TSX (TypeScript + JSX) |
 | `tree-sitter-javascript.wasm` | JavaScript             |
+| `tree-sitter-c-sharp.wasm`    | C#                     |
 
 ## Lockfile
 
--- a/grammars/grammars.lock
+++ b/grammars/grammars.lock
@@ -56,6 +56,13 @@
       "package": "tree-sitter-javascript",
       "version": "0.23.1",
       "sha256": "4a378293fe7853cbee2836023be072dafa0e53b3b5edb245920838ca834ed121"
+    },
+    {
+      "file": "tree-sitter-c-sharp.wasm",
+      "source": "npm",
+      "package": "tree-sitter-c-sharp",
+      "version": "0.23.1",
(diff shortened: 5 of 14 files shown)


stasson:
## Self-review (from `/review` per AGENTS.md workflow)

**Branch:** `worktree-feat-csharp-grammar` → `main`
**Scope:** +498 / -11 across 11 files

### Overview

Adds C# (`.cs`) to the source-parser dispatch via `tree-sitter-c-sharp@0.23.1`. Mechanically follows the `LANGUAGE_SPECS` pattern. Non-mechanical decisions are version pin (ABI v14 vs v15), `urlFile` field on `NpmGrammar` (upstream underscore vs on-disk hyphen), and `csharpItemName` with three exception categories.

### Strengths

- Surgical, additive change. Each touched file gains one row or one line. Existing five grammars are untouched.
- `urlFile` design is minimally invasive: optional field, used only by the C# entry, default-falls-back via `?? file`.
- Probe-verified deviations are documented inline: destructor non-special-casing (`language_spec.ts:144-149`), `record_declaration` covering `record struct` too (`:286`), field/event-field walk into `variable_declarator` (`:138-170`).
- Test discipline: every test asserts `entry.source.kind === "doc-comment"` BEFORE the narrowing block, matching the convention at `source_test.ts:893`. A parser regression to a different `EntrySource` shape would fail the assertion instead of silently skipping it.
- `entries.length === 14` pin in the coverage matrix (`source_test.ts:988`) catches wholesale extraction regressions in addition to per-row assertions.
- All 16 node types in the new `enclosingItemTypes` list are exercised by tests. Negative-case (`//`) and Javadoc-fallback (`/** */`) both covered.

### Code quality observations

- `csharpItemName` doc comment says "Three exceptions" but lists 4 node types across 2 categories (`language_spec.ts:139`). Minor wording inaccuracy; defer.
- Carrier-class workaround (`source_test.ts:906-910`, `_NamespaceCarrier`) is a structural artifact of how the sibling walker handles the namespace's first inner comment. The inline comment explains it. Namespace attribution itself is independently covered by the `traditional namespace name` test (`:1015-1031`).
- `attributeSkipTypes: ["comment", "attribute_list", "modifier"]` is defensive — probe confirmed `modifier` and `attribute_list` sit inside declarations (not as preceding siblings) in v0.23.1. Listed for safety against future grammar reshuffles.
- `urlFile` is npm-only (`GithubGrammar` doesn't carry the field). Not relevant today; if a GitHub-sourced grammar ever needed it, the field would have to be lifted. No action.

### Risks

- **Grammar ABI lock-in.** Pinning to 0.23.1 freezes C# at C# 11/12 syntax surface. When `web-tree-sitter` is upgraded (separate effort), C# can move to 0.23.5+. Documented in the GRAMMARS entry comment.
- **No fixture-extension exclude needed.** `deno.json` `fmt.include` is TS/JS-only; `dprint.json` `includes` is `.md`/`.json`/`.yaml`/`.yml`/`.toml`. `.cs` is invisible to both.
- **JS-family PR conflicts.** Mechanical, additive. Merge order coordinated at PR time.

### Security

- WASM grammar via Git LFS — `.gitattributes` `grammars/*.wasm filter=lfs` covers the new file automatically. SHA-256 in `grammars.lock:43` matches the LFS pointer oid in `grammars/tree-sitter-c-sharp.wasm:2`. Tamper-evident.
- No new npm runtime dependency in `packages/markspec/deno.json`.

### Test coverage

10 new `Deno.test` cases, full source_test.ts suite passes 44/44. Tests use deterministic source strings, look up by `displayId` not by index, and assert finite cardinality. No flakiness risk.

### Performance

No implications. WASM grammar lazy-loads only when a `.cs` file is parsed, via the existing `loadGrammar()` cache (`grammars.ts:55-67`).

### Assessment

**Approved.** Clean implementation following established patterns, well-documented deviations, disciplined tests with correct type-narrowing, and explicit version-pin rationale. The "Three exceptions" wording in the `csharpItemName` doc comment is the only nit and is non-blocking.

Issue #518, opened 2026-05-26
story: formal $Identifier entity-resolution model (M050/M051, typed identifiers, profile Aliases)


stasson:
## Formal `$Identifier` entity-resolution model

Parent epic: #500
Follow-up to: #502

### Motivation

#502 shipped the pragmatic minimal version of the Q500 `$Identifier` resolver leg — a corpus-scan that aggregates every `entity-ref` body token into a `Set<string>` and silences Q500 on phrases that match any bare name observed anywhere in the workspace. This implements [ADR-021 Decision 1](https://github.com/driftsys/markspec/blob/main/docs/architecture/adr-021-prose-analysis-flagship-build.md)'s additive-enrichment invariant but does not yet implement the formal entity-resolution semantics from `core-data-model §2.5.2` / `§4.6` ([MSL-M050](https://markspec.dev/lint/rules/msl-m050) / [MSL-M051](https://markspec.dev/lint/rules/msl-m051)).

The pragmatic minimal version cannot distinguish:

- A `$Identifier` that is **declared** somewhere (definition site) from one that is **only used** (no definition anywhere).
- A typed identifier (`$BrakePressure: Force`) from an untyped one.
- A profile-`Aliases`-resolved identifier from one that resolves only through the entity registry.

Each of these distinctions is needed to ship the deferred MSL-M050 ("undefined identifier") and MSL-M051 ("untyped identifier") rules.

### Dependencies

**Blocked on:** nextgen entity-resolution spec on `main` (`core-data-model §2.5.2` / `§4.6`). Per [ADR-014 §Decision-4](https://github.com/driftsys/markspec/blob/main/docs/architecture/adr-014-canonical-body-ast.md), MSL-M050 / MSL-M051 are deferred-by-dependency on this spec landing.

### Scope

- Define the `$Identifier` definition-site grammar (where does a declaration live? In a typl block? In a sibling attribute?).
- Build a registry that distinguishes definitions from uses across the corpus.
- Implement MSL-M050 (undefined identifier).
- Implement MSL-M051 (untyped identifier) if typing is in scope.
- Wire profile `Aliases` resolution ([`listing-directives §4.3 R4-g`](https://github.com/driftsys/markspec/blob/main/docs/spec/internal/markspec-listing-directives.md)) into the same registry.
[...]

PR #519, opened 2026-05-26
feat(lint): corpus-scan $Identifier resolver hook for Q500

### LINKED ISSUE #502: $Identifier registry resolver hook for MSL-Q500 (blocked on ADR-016 marker pass)
## `$Identifier` registry resolver hook for MSL-Q500

Parent epic: #500

### Motivation

[ADR-021 Decision 1](https://github.com/driftsys/markspec/blob/main/docs/architecture/adr-021-prose-analysis-flagship-build.md) ships `MSL-Q500 xref-glossary-undefined` with a no-op `IsIdentifierHook` for the `$Identifier` registry leg of the resolver. The hook exists at the contract level so the ADR-016 marker pass can plug into it without rule changes.

Today the hook always returns `false`. This means Q500 fires on capitalized phrases that overlap with `$Identifier` entity references — e.g. `BrakePedalPressed` (a Pascal-case entity ref the marker pass would resolve via the symbol registry). The behaviour is correct under the deferred-resolver posture (`MSL-M050`/`MSL-M051` from core §4.6 are themselves deferred-by-dependency), but as the marker pass lands, those tokens should silently resolve and stop firing Q500.

ADR-021's behavioural invariant: **more tokens resolve as the resolver grows → Q500 fires *less*, never *more***. The hook is the seam that preserves the invariant.

### Goal

When the ADR-016 marker pass and `$Identifier` registry land, replace the no-op `IsIdentifierHook` in [`core/lint/rules/xref.ts`](https://github.com/driftsys/markspec/blob/main/packages/markspec/core/lint/rules/xref.ts) with a callback that consults the registry.

Mechanically: thread the registry (probably accessible from `Entry.bodyTokens` per ADR-016 or from a separate marker-pass output) into `runLint`'s `LintOptions`, build a closure inside `runner.ts`, pass it to `runXrefRules`. The existing test `Q500: silent for $Identifier leg via no-op hook` (slice 5) already pins the contract; new tests will cover the integration.

### Dependencies

**Blocked on:** ADR-016 marker-pass implementation. Issue [#409](https://github.com/driftsys/markspec/issues/409) is the closed epic for the body-token AST; the marker pass plumbing landed as part of that, but the `$Identifier` registry / symbol-resolution model (referenced from `core-data-model §2.5.2` / `§4.6`, `MSL-M050`/`M051`) is deferred-by-dependency per ADR-014 §Decision-4 — it needs the nextgen entity-resolution spec on `main` first.

When the registry exists, this issue is a small PR (~30 lines + tests). No design needed — ADR-021 Decision 1 pinned the integration shape already.

### Out of scope

- The marker pass / entity-registry itself (separate epic).
- Profile `Aliases` resolution (`listing-directives §4.3 R4-g`) — also deferred, but separable from this story.

### References

- [ADR-021 Decision 1](https://github.com/driftsys/markspec/blob/main/docs/architecture/adr-021-prose-analysis-flagship-build.md)
- [`core/lint/rules/xref.ts`](https://github.com/driftsys/markspec/blob/main/packages/markspec/core/lint/rules/xref.ts) — `IsIdentifierHook` definition + current no-op
- ADR-014 Decision-4 — `MSL-M050`/`M051` deferred-by-dependency status

stasson:
## Summary

- Replaces Q500's no-op `IsIdentifierHook` with a corpus-scan closure built from every entry's `entity-ref` body tokens (ADR-021 Decision 1's pragmatic minimal version).
- Adds `buildIdentifierIndex` + `buildIsIdentifierHook` exports in [`core/lint/rules/xref.ts`](packages/markspec/core/lint/rules/xref.ts) so the seam is discoverable for the follow-up story.
- Closes #502. Formal MSL-M050/M051 entity-resolution semantics remain deferred — tracked in #518.

## Behavioural contract

ADR-021 Decision 1's additive-enrichment invariant: **Q500 fires *less* as the resolver grows, never *more*.** The corpus-scan closure honors this — it only adds members to the silenced set, never removes them. Reviewers should sanity-check this on the diff.

## Test plan

- [x] `deno test packages/markspec/core/lint/rules/xref_test.ts` — 3 new unit tests for `buildIdentifierIndex` (aggregation, empty input, non-entity-ref token filtering)
- [x] `deno test packages/markspec/core/lint/runner_test.ts` — 2 new integration tests (corpus-scan silences cross-entry $Identifier; unknown phrases still fire)
- [x] `just build` — all 2657 tests pass, lint + type-check + compile clean
- [x] `deno fmt --check && dprint check` — format gates clean

### DIFF
--- a/packages/markspec/core/lint/rules/xref.ts
+++ b/packages/markspec/core/lint/rules/xref.ts
@@ -13,12 +13,57 @@ import type { GlossaryIndex } from "../glossary.ts";
 import { deriveTermSlug } from "../../parser/glossary.ts";
 import { offsetToRange } from "../range_util.ts";
 
-/** Hook signature for the deferred $Identifier registry leg. Returns
- * true when the token is a resolvable $Identifier (i.e. should be
- * skipped by Q500). Today, always returns false; ADR-016's marker
- * pass replaces this with a real registry lookup. See ADR-021
- * Decision 1. */
+/** Hook signature for the $Identifier registry leg. Returns true when
+ * the token is a resolvable $Identifier (i.e. should be skipped by
+ * Q500). See ADR-021 Decision 1 for the integration seam, and
+ * {@linkcode buildIdentifierIndex} + {@linkcode buildIsIdentifierHook}
+ * for the corpus-scan implementation that backs it today. */
 export type IsIdentifierHook = (token: string) => boolean;
+
+/**
+ * Build a corpus-wide `$Identifier` index from every entry's `bodyTokens`.
+ *
+ * Aggregates the bare name (text with leading `$` stripped) of every
+ * `entity-ref` token across `entries` into a `Set<string>`. The set is
+ * the data backing the {@linkcode IsIdentifierHook} closure produced by
+ * {@linkcode buildIsIdentifierHook}.
+ *
+ * Behavioural contract (ADR-021 Decision 1, additive-enrichment invariant):
+ *   - Every name in the index is observed in the corpus as `$Name`.
+ *   - The index is purely additive: more entries → more silenced phrases,
+ *     never more diagnostics.
+ *
+ * Scope note: this is the pragmatic minimal resolver leg per issue #502.
+ * The formal `MSL-M050`/`MSL-M051` entity-resolution model (definition
+ * vs. use, typed identifiers, profile Aliases) is deferred to a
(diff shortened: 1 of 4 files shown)


stasson:
Re-triggering CI — workflows did not fire on initial open (suspect webhook drop; sibling PR #520 opened seconds later got CI fine).

PR #559, opened 2026-05-28
fix(lsp): prevent duplicate display IDs across rapid scaffold accepts

### LINKED ISSUE #5: `core/` library module and public exports
Epic: driftsys/board#13

## Context

Create `core/mod.ts` as the public API surface. Set up `deno.json`
with exports map: `"./core": "./core/mod.ts"`. The module exports
the parser, model types, format, validate, and compile functions
(stubs initially, filled by later stories). An external project can
`import { parse } from "@driftsys/markspec/core"`.

## Acceptance Criteria

- [ ] A separate Deno project imports `@driftsys/markspec/core` and calls `parse()` on a Markdown string
- [ ] Types are exported correctly
- [ ] `deno.json` exports map includes `"./core": "./core/mod.ts"`

stasson:
## Summary

Devex item **#5** of the editing-experience backlog: the debounce-window race that produces duplicate display IDs.

The per-file parse that feeds `WorkspaceIndex` is debounced (50ms) and async, so the index lags the document. When an author accepts a scaffold completion, the inserted entry does not reach `byDisplayId` until the next parse fires. If a second scaffold accept lands inside that window, both `completionItem/resolve` calls invoke `getNextDisplayIdNumber`, both observe the same indexed maximum, and both mint the **same** display ID — a duplicate the validator flags on the following debounce.

### Is the race real? (it is — and it's not what "race" usually means)

I confirmed the mechanism is live, not dead code. `onCompletionResolve` is **synchronous**, so two resolves do serialize on the event loop and never interleave. The bug is therefore a **staleness** race, not a concurrency race: neither resolve sees the other's freshly-inserted entry because the debounced parse that would add it to the index has not run yet. The integration test reproduces exactly this — a static index whose maximum never moves between the two resolves — and shows both mint the same number without the fix, distinct numbers with it.

## Mechanism

| File | What |
|---|---|
| `lsp/id_reservations.ts` *(new)* | Short-lived in-memory reservation set keyed by `(prefix, suffix)`. `reserve` / `isReserved` / `release` / `reservedNumbersFor`, plus `mintReservedNumber` (atomic compute-and-reserve, index lookup injected as a callback to keep the module decoupled from `WorkspaceIndex`). 60s TTL evicts reservations from abandoned accepts. |
| `lsp/workspace.ts` | `getNextDisplayIdNumber` gains an optional `reserved: ReadonlySet<number>` param folded into the maximum. Backwards-compatible — existing callers pass nothing. |
| `lsp/server.ts` | `onCompletionResolve` mints + reserves atomically *before* rendering the snippet, so an immediately-following resolve sees the number as taken. After each edit-driven `parseAndUpdateFile` (`onDidChangeContent`, `onDidOpen`), `releaseObservedReservations` drops reservations now present in the index. |

### Deliberately unchanged
- **Build-time scaffold path** (`onCompletion` / `getEntryTypes`) does **not** reserve — it is best-effort; the resolve-time path is authoritative. Reserving at build time would over-reserve when authors merely browse the list.
- **Parse debounce** is untouched. The debounce exists for performance under heavy typing; the race is fixed by reservation, not by removing it.
- Reservations live for the LSP process lifetime only. On restart the parsed index is authoritative again.

## Test plan

- [x] `deno fmt --check` — clean
- [x] `deno lint` — clean
- [x] `deno test --allow-read --allow-write --allow-run --allow-env --allow-ffi` — **2958 passed, 0 failed, 3 ignored** (+12 new)
- [x] `deno check` (all entry points) — clean
- [x] Unit tests for `id_reservations.ts`: reserve, isReserved, release (incl. unknown no-op), `(prefix, suffix)` keying independence, `reservedNumbersFor`, TTL eviction, `_reset`
- [x] `workspace_test.ts`: `getNextDisplayIdNumber` skips a reserved number; reserved-max dominates; empty set is a no-op
- [x] Integration: two back-to-back mints against a **stale** real `WorkspaceIndex` render distinct display IDs (`STK_0003` then `STK_0004`) via the exact `mintReservedNumber` + `getNextDisplayIdNumber` + `renderScaffoldSnippet` composition `server.ts` uses

## Backlog status

- [x] #1 display-ID pattern share — shipped (#549)
- [x] #2 relation-target filtering — shipped (#554)
- [x] #3 server-side prefix filter — shipped (#556)
- [ ] #4 mid-typed `[STK_AEB_…` auto-complete next sequential number — **still open** (not yet started)
- [x] **#5 debounce-window duplicate-ID race — this PR**

> Note: my brief assumed #4 would land before #5, but #4 has not shipped. This PR closes #5 only; the backlog is **not** fully closed until #4 lands.

### DIFF
--- a/packages/markspec/lsp/id_reservations.ts
+++ b/packages/markspec/lsp/id_reservations.ts
@@ -1,0 +1,151 @@
+/**
+ * @module lsp/id_reservations
+ *
+ * Short-lived, in-memory reservation set for display-ID numbers handed
+ * out by the scaffold-completion resolve handler but not yet observed in
+ * the parsed workspace index.
+ *
+ * ## Why this exists
+ *
+ * The per-file parse that feeds {@linkcode WorkspaceIndex} is debounced
+ * (50 ms) and asynchronous, so the index lags the document. When an
+ * author accepts a scaffold completion, the inserted entry does not reach
+ * `byDisplayId` until the next parse fires. If a second scaffold accept
+ * lands inside that window, both `completionItem/resolve` calls invoke
+ * `getNextDisplayIdNumber`, both observe the same indexed maximum, and
+ * both mint the *same* display ID — a duplicate the validator flags on
+ * the following debounce.
+ *
+ * This is a *staleness* race, not an event-loop concurrency race: the
+ * resolve handler is synchronous, so two resolves serialize, but neither
+ * sees the other's freshly-inserted entry because the parse has not yet
+ * run. {@linkcode reserve} records a number the instant resolve hands it
+ * out so the next resolve skips it; {@linkcode release} is called once a
+ * parse observes the entry in the index.
+ *
+ * Reservations live for the LSP process lifetime only — on restart the
+ * parsed index is authoritative again. A {@linkcode RESERVATION_TTL_MS}
+ * timeout drops reservations for completions abandoned mid-accept so a
+ * number is never blocked forever.
+ */
+
+/** Reservations older than this are evicted lazily on read. */
+const RESERVATION_TTL_MS = 60_000;
+
+/**
+ * Bucket key for a `(prefix, suffix)` pair. The NUL separator can never
+ * appear in a display-ID prefix or suffix, so it unambiguously joins the
+ * two halves without collisions (e.g. `("A_", "B")` vs `("A", "_B")`).
+ */
+function keyOf(prefix: string, suffix: string): string {
+  return `${prefix}\0${suffix}`;
+}
+
+/** key → (reserved number → reservation timestamp in ms). */
+const reservations = new Map<string, Map<number, number>>();
+
+/** Shared empty result for the no-bucket case — never mutated. */
+const EMPTY: ReadonlySet<number> = new Set<number>();
+
+/** Wall-clock source; overridable in tests via {@linkcode _setNow}. */
+let nowFn: () => number = () => Date.now();
+
+/**
+ * Drop entries in `bucket` whose reservation is at least
+ * {@linkcode RESERVATION_TTL_MS} old. Mutates the bucket in place.
+ */
+function evictStale(bucket: Map<number, number>, now: number): void {
+  for (const [n, ts] of bucket) {
+    if (now - ts >= RESERVATION_TTL_MS) bucket.delete(n);
+  }
+}
+
+/**
+ * Record that display-ID `n` (for the given `prefix`/`suffix` pattern)
+ * has been handed out. Overwrites the timestamp if `n` was already
+ * reserved, refreshing its TTL.
+ */
+export function reserve(prefix: string, suffix: string, n: number): void {
+  const key = keyOf(prefix, suffix);
+  let bucket = reservations.get(key);
+  if (!bucket) {
+    bucket = new Map();
+    reservations.set(key, bucket);
+  }
+  bucket.set(n, nowFn());
+}
+
+/**
+ * Return `true` when `n` is currently reserved for the given pattern.
+ * Evicts stale reservations first, so a number past its TTL reads as
+ * not reserved.
+ */
+export function isReserved(prefix: string, suffix: string, n: number): boolean {
+  const bucket = reservations.get(keyOf(prefix, suffix));
+  if (!bucket) return false;
+  evictStale(bucket, nowFn());
+  return bucket.has(n);
+}
+
+/**
+ * Drop the reservation for `n`. Called once a parse observes the entry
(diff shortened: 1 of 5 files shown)


stasson:
## Code review — `/code-review high --comment`

Reviewed `origin/main...HEAD` (5 files, +452/−2) with 7 finder angles (3 correctness, 3 cleanup, 1 altitude) + 1-vote verification.

**No correctness bugs found.** The decomposition digit-guard, TTL boundary (`>=` at exactly 60 000 ms), integer round-trip (`parseInt` of a zero-padded segment), `break` scoping, and the cold-start no-leak (no reservations exist before the first parse) all verified correct. The findings below are all quality/cleanup — none block merge.

### 1. Reuse — display-ID decomposition is now duplicated a third time *(minor)*
[lsp/server.ts:359](packages/markspec/lsp/server.ts#L359) `releaseObservedReservations` decomposes an ID with the same `startsWith` / `endsWith` / `slice` / `parseInt` shape already implemented in [workspace.ts `getNextDisplayIdNumber`](packages/markspec/lsp/workspace.ts#L191) and [display_id.ts `highestDisplayIdNumber`](packages/markspec/core/profile/display_id.ts#L84-L101). Three copies of the same parse is a maintenance/bug surface. **Suggestion:** extract `decomposeDisplayId(id, shape): number | undefined` into `core/profile/display_id.ts` and call it from all three sites.

### 2. Efficiency — release path does work on every keystroke even when nothing is reserved *(minor)*
[lsp/server.ts:352](packages/markspec/lsp/server.ts#L352) `releaseObservedReservations` runs after every 50 ms debounced parse (`onDidChangeContent`). It calls `getScaffoldPatterns()` (regex-parses every profile type's `display-id-pattern`) and then scans `entries × patterns` — but the reservation map is empty in the overwhelmingly common case (no scaffold accept in flight). **Suggestion:** export `hasAnyReservations(): boolean` from `id_reservations.ts` and early-return from `releaseObservedReservations` when it's false, skipping the pattern parse + entry scan entirely.

### 3. Minor — read-path TTL eviction leaves empty buckets
[lsp/id_reservations.ts](packages/markspec/lsp/id_reservations.ts) `isReserved` / `reservedNumbersFor` call `evictStale` but, unlike `release`, don't drop a bucket that becomes empty. Bounded by the number of distinct `(prefix, suffix)` pairs (≈ profile type count, tiny), so not a growing leak — noting for symmetry only. Optional.

### Verified as non-issues (not changing)
- `mintReservedNumber`'s `nextFree` callback is a deliberate decoupling seam (keeps `id_reservations` independent of `WorkspaceIndex`) and is what the integration test exercises — keeping it.
- NUL (`\0`) bucket key is documented and safe (cannot appear in a prefix/suffix).
- Build-time scaffold path intentionally does not reserve; only the displayed label can be momentarily stale, never the inserted text.

I'll fold #1 and #2 into this PR (both are small and on the touched code).

stasson:
**Follow-up — review findings #1 and #2 folded in** (commit `9925546`):

- **#1 (reuse):** extracted `decomposeDisplayId` into `core/profile/display_id.ts` and routed `highestDisplayIdNumber`, `getNextDisplayIdNumber`, and the reservation-release path through it. The helper is **lenient by default** to preserve the documented next-ID contract (`highestDisplayIdNumber` counts `STK_AEB_0099_extra` → 99 so the next ID never collides), with a `strict` opt-in for the release path that needs prefix-sharing patterns (`STK_{n:4d}` vs `STK_{n:4d}-draft`) to disambiguate. No behavior change to either next-number computer.
- **#2 (efficiency):** added `hasAnyReservations()` early-out so the per-keystroke release scan skips the pattern parse + entry walk entirely when nothing is in flight (the common case).
- **#3** (empty-bucket eviction) left as-is — bounded by profile-type count, negligible.

`deno fmt --check` / `deno lint` / `deno check` clean; full suite **2964 passed, 0 failed, 3 ignored**.

stasson:
**Correction + post-merge verification.** My earlier note said item #4 was still open — that was true when I checked mid-implementation, but #4 (PR #558) merged 16 s before this PR, so the **full 5-item editing-experience devex backlog is now complete**.

Both #558 and #559 rewrote the same `onCompletionResolve` handler and `ScaffoldCompletionData`, and #559 was branched before #558 landed — so I verified the integrated result on `main` (`a157e9c`): the handler now correctly carries **both** the atomic reservation mint (this PR) and #558's `replacementRange` textEdit rebuild. `deno check` clean; LSP + display-ID suites **354 passed, 0 failed**. No bad interaction from the near-simultaneous merge.

PR #569, opened 2026-05-30
feat(core): model interfaces as contracts, not components

stasson:
## Summary

Re-parents `SoftwareInterface` and `HardwareInterface` from `Component` to `Contract` (the `Specification` family). An interface is a **specification of a boundary**, not a building block — the prior `Component` parentage was a modeling error (interfaces reused only 1 of `Component`'s 6 attributes, and the provider relations pointed backwards). UML / SysML / AUTOSAR / ICD all treat an interface *definition* as a contract distinct from the component that provides or requires it.

- **type_hierarchy:** interfaces now extend `Contract`; `HardwareInterface` keeps its physical attrs (`Bus-protocol`, `Connector-type`, `Voltage-level`, `Signal-direction`). Names kept (literature terms).
- **compiler:** `Provides`/`Requires` wired as `provides`/`requires` traceability links (`ATTR_TO_LINK_KIND` + `KNOWN_LINK_KINDS`) — symmetric provider/consumer relations Component → Contract.
- **attributes:** add `Depended-on-by` generated inverse, freeing `Required-by` to be the inverse of `Requires`.
- **lsp:** register `Provides`/`Requires` in the trace-keyword lists (completions + context) so they complete and are recognised like other trace keys.
- **docs(spec):** interface types moved under `Contract`; relations + attributes tables updated (`Depends-on`'s inverse renamed to `Depended-on-by`).

`Allocated-to` is unchanged (not reused for provider-hosting); interface discipline stays type-driven. No migration tooling (pre-1.0). An `Element` abstract parent over `{Component, Unit}` (ASPICE-4.0 alignment) is intentionally deferred to a later phase.

## Test Plan

- [x] `just check` — 3096 passed, 0 failed (lint + test + type-check, `--allow-net`)
- [x] `just compile` — binary builds
- [x] `deno fmt --check` and `dprint check` — clean
- [x] New tests: type-hierarchy inheritance (interfaces inherit Contract chain, not Component), `provides`/`requires` link emission, MSL-T022 reflects the re-parent, LSP `Provides`/`Requires` keyword triggers
- [x] `markspec validate docs/product/*.md` — exit 0

### DIFF
--- a/docs/spec/internal/markspec-core-data-model.md
+++ b/docs/spec/internal/markspec-core-data-model.md
@@ -115,8 +115,8 @@ The information type is expressed by a single `Type:` attribute (ADR-004 §Part
   TitleCase. Always available regardless of profile.
 - **Core concrete types** — the 15 built-in subtypes (ADR-003 §Part 1):
   - Under `Specification`: `Requirement`, `Test`, `Contract`, `Record`, `Risk`
-  - Under `Component`: `SoftwareComponent`, `HardwareComponent`,
-    `SoftwareInterface`, `HardwareInterface`
+  - Under `Component`: `SoftwareComponent`, `HardwareComponent`
+  - Under `Contract`: `SoftwareInterface`, `HardwareInterface`
   - Under `Unit`: `SoftwareUnit`, `HardwareUnit`
   - Standalone under `Item`: `Definition`
 - **Profile-declared concrete types** — extend any core type by inheritance.
@@ -241,8 +241,8 @@ parser knows which attribute keys are core-defined vs profile-declared.
 | `Component`         | `Kind`, `Part-of`, `Realizes`, `Depends-on`, `Provides`, `Requires`   | §Part 2 “Component"         |
 | `SoftwareComponent` | `License`, `Build-manifest`, `Package-manager`                        | §Part 2 “SoftwareComponent" |
 | `HardwareComponent` | `Manufacturer`, `Part-number`, `Datasheet`                            | §Part 2 “HardwareComponent" |
-| `SoftwareInterface` | _(inherits Component; no additions in core)_                          | §Part 2 “SoftwareInterface" |
-| `HardwareInterface` | `Bus-protocol`, `Connector-type`, `Voltage-level`, `Signal-direction` | §Part 2 “HardwareInterface" |
+| `SoftwareInterface` | _(inherits Contract → Specification; no additions in core)_           | §Part 2 “SoftwareInterface” |
+| `HardwareInterface` | `Bus-protocol`, `Connector-type`, `Voltage-level`, `Signal-direction` | §Part 2 “HardwareInterface” |
 | `Unit`              | `Part-of`, `Realizes`, `Depends-on`                                   | §Part 2 “Unit"              |
 | `SoftwareUnit`      | `Source`, `Symbol`, `Language`                                        | §Part 2 “SoftwareUnit"      |
 | `HardwareUnit`      | `Manufacturer`, `Part-number`, `Datasheet`, `Footprint`, `Value`      | §Part 2 “HardwareUnit"      |
--- a/docs/spec/internal/markspec-listing-directives.md
+++ b/docs/spec/internal/markspec-listing-directives.md
@@ -64,10 +64,11 @@ separately and the spec does not unify the three.
 **Cross-references to the companion spec:**
 
 - The Component types a `markspec:components` listing admits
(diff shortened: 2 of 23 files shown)


stasson:
## Code review — findings (high-effort, verified)

Automated review (per AGENTS.md) of the interface-as-contract change. The core re-parent + link wiring is correct, but the review found that this codebase keeps **several independent hardcoded lists of trace relations / type families**, and the PR updated only the link-production one (`ATTR_TO_LINK_KIND`). The others are now inconsistent with the re-parent. None are crashes; all are correctness/consistency regressions worth fixing before merge.

### Code (confirmed)

1. **`core/validator/trace_types.ts:37,53` — `Tests` / `Affects` no longer accept interface targets (regression).**
   Both allow `[Component, Unit]`. `isTargetTypeCompatible` walks the parent chain; before the re-parent `SoftwareInterface`/`HardwareInterface` reached `Component`, so `Tests: <interface>` / `Affects: <interface>` validated. After re-parenting to `Contract` they walk `Contract → Specification → Item` and never hit `Component`, so those now emit a **false `MSL-R083`**. Decide intended behavior: under the new model a test should `Verifies:` an interface (a `Contract`) — which works — so this may be acceptable, but it is a **silent semantic change** that should be made explicit (and, if interfaces should stay testable, add `Contract` to those target sets).

2. **`core/validator/trace_types.ts:41-46` — `Provides`/`Requires` target set should be `[Contract]`, not `[SoftwareInterface, HardwareInterface]`.**
   The relations spec (this PR's `relations.md`) defines the target as `Contract`. With the current narrow set, `Provides:`/`Requires:` to a plain `Contract` (or a profile subtype such as `API`/`ICD`) is wrongly rejected by `MSL-R083`. Because `isTargetTypeCompatible` walks ancestors, `[Contract]` accepts `Contract` **and** both interface subtypes — matching the docs.

3. **`core/lock/resolve.ts:212` — lockfile `TRACE_KEYS` omits `Provides`/`Requires`.**
   `extractEdgeQuads` has its own trace-key list (independent of `ATTR_TO_LINK_KIND`); it still lists only the original 11 keys, so `Provides`/`Requires` edges are excluded from the canonical edge hash. `markspec lock` / drift detection is blind to provides/requires relationship changes.

4. **`core/validator/listing.ts:53-58` — `COMPONENT_FAMILY` still includes the interface types.**
   `MSL-L043` (components-listing validation) still treats `SoftwareInterface`/`HardwareInterface` as component-family, so a `markspec:components` listing silently accepts interface (now spec) entries. Decide: remove them from the set and update the error message (~:283).

### Docs (stale — not updated by the PR)

5. **`docs/spec/internal/markspec-core-data-model.md:244`** — `SoftwareInterface` row still says "inherits Component".
6. **`docs/spec/internal/markspec-listing-directives.md:67`** — components listing still documented as admitting the interface types (pairs with #4).
7. **`docs/spec/model/annex-profile-schema.md:144`** — interfaces still listed under "Component subtypes" as `extends:` targets.
8. **`docs/spec/model/attributes.md:194`** — `HardwareInterface` profile-attrs cell lists only `Bus-protocol`, `Voltage-level`; its declared own-attrs also include `Connector-type`, `Signal-direction`.

### Considered and dropped
- `docs/architecture/adr-017` tree still shows interfaces under Component — ADRs are immutable-once-accepted; the tree reflects state at writing.
- `render/styles/mod.ts` ID monospacing — pre-existing for *all* link attrs, and the module is not wired into the active render pipeline.

### Root cause / altitude
Findings 1-4 share one cause: trace-relation metadata is duplicated across `ATTR_TO_LINK_KIND` (compiler), `TRACE_RULES` (validator), `TRACE_KEYS` (lock), and `COMPONENT_FAMILY` (listing). A follow-up could derive these from a single source so a relation/type change cannot half-land like this.

stasson:
Findings addressed in b2366ee (code+tests) and 124a343 (docs):
- 1 Tests/Affects now include Contract so interface specs stay valid targets (no false MSL-R083).
- 2 Provides/Requires target set changed to [Contract] (accepts Contract + all subtypes via the hierarchy walk).
- 3 lockfile TRACE_KEYS now includes Provides/Requires.
- 4 COMPONENT_FAMILY drops the interface types; MSL-L043 now flags them in components listings (error message + listing-directives doc updated).
- 5-8 stale docs (core-data-model, listing-directives, annex-profile-schema, attributes) aligned with the Contract re-parent.

New/updated tests: trace_types_test (4), resolve_test (2), validate_listing e2e (1). just check green (lint + test + type-check). Root-cause note stands: the four duplicated relation/family lists are a candidate for a single-source follow-up.

Issue #582, opened 2026-05-31
Profile: first-class authoring-stance + document location/filename metadata


stasson:
## Problem

Profiles such as `@ampere/seed` (Renault **FastTrack**) need to express, **per entry type** and **per document type**, *how that artifact is meant to be produced*:

- **hand-authored** Markdown entry,
- **colocated-in-code** (a doc-comment MarkSpec entry whose test function is the test),
- **generated** from build config / manifests / IDL (e.g. `software-api`, `sw-component`).

For documents they also need the recommended **on-disk location** and **filename template**.

Today the profile schema (core-schema 1, markspec 0.7.0) rejects any key outside the known set:

- adding `authoring:` to a type → `error[PROFILE-TYPE-005]: unknown key 'authoring'`
- adding `path:` / `filename-template:` to a document → `error[PROFILE-LOAD-003]: unknown key ...`

So this guidance can currently only live in free-text `description` prose.

This matters most for **AI agents**. An agent authoring requirements reads the profile (via `markspec profile describe` and `markspec compile --format json`) to decide what to do. With the stance buried in prose, the agent cannot reliably tell that `software-api` / `sw-component` are *generated, do-not-hand-write* types, nor where a `SyRS` should be written. **Structured fields make the profile self-describing to tooling and agents.**

## Proposed solution

Add optional, validated profile fields:

**Types**
- `authoring: generated | colocated-in-code | hand-authored` (optional; default `hand-authored`)
- `authoring-note:` (optional free text — the source/exception, e.g. "manual-test fallback authored as Markdown")

**Documents (`documents.types[]`)**
- `authoring:` (same enum)
- `path:` — recommended directory, e.g. `specs/system/`
- `filename-template:` — e.g. `{scope}.syrs.md` (same placeholder grammar as display-id patterns)

**Surface them in**
- `markspec profile describe type|document …`
- `markspec compile --format json` (so agents / codegen consume them structurally)
- optionally a `markspec check` advisory when an entry of a `generated`/`colocated-in-code` type is found hand-written as Markdown.

## Alternatives considered

- **Prose stem in `description`** — the current interim in `@ampere/seed`: greppable but unvalidated, absent from the JSON model, and easy for an agent to miss. Adopted now as a stopgap; this story is the durable fix.
- **Keep the guidance only in the book / skills** — separates it from the machine-readable profile the agents actually load; the profile stays silent on intent.

## Context

[...]

PR #595, opened 2026-06-05
feat(core): classify named types via counter-less display-id-pattern

### LINKED ISSUE #594: Allow type inference for named (non-numbered) types via counter-less display-id-pattern
## Summary

markspec classifies an Authored entry's type two ways: an explicit `Type:`
trailer, or a `display-id-pattern` match. But `compileDisplayIdPattern`
requires **exactly one `{n}` numeric counter**, so a type whose IDs are
*named, not numbered* — e.g. components `SWC_LIGHT_CTRL`, `HWC_PIU` — cannot
declare a pattern, and therefore cannot be classified by prefix. Every such
entry must hand-write an explicit `Type:`.

**Request:** allow a **counter-less `display-id-pattern`** (literal prefix +
named segment, no `{n}`) so named types are classified by prefix — e.g.
`SWC_{name}` → `sw-component`, `HWC_{name}` → `hw-component`.

## Current behavior (verified on v0.7.2)

- `compileDisplayIdPattern` throws on any pattern without a counter
  (`packages/markspec/core/validator/pattern.ts` — grammar
  `pattern := segment+ (exactly one counter)`; the `counters === 0` branch
  throws `missing {n} placeholder`).
- A named type therefore declares *no* pattern, so `classifyEntry`'s pattern
  loop (`packages/markspec/core/validator/types.ts`, step 2) never matches it,
  and the entry is unclassified unless `Type:` is explicit (`MSL-T003` in
  strict mode).
- Reproduced with the SEED profile (`@ampere/seed`): an `SWC_LIGHT_CTRL`
  entry with no `Type:` resolves to `<UNRESOLVED>`; adding
  `display-id-pattern: "SWC_{name}"` errors with `missing {n} placeholder`.
- Late-stage inference (steps 5/6/8, `types.ts` `inferTypeFromDisplayIdShape`
  / `inferTypeFromLateStageChain`) only resolves the abstract *core* type and
  emits an `MSL-T021` warning telling the author to declare `Type:` anyway —
  it does not assign the profile type, and it keys off attributes *on* the
  entry, not the relation pointing at it.

## Impact

Profiles with named component/element types must hand-write `Type:` on every
such entry, even though the prefix already identifies the type unambiguously.
In SEED, `sw-component` and `hw-component` are deliberately *"named, not
numbered (SWC_DSG, SWC_LIGHT_CTRL)"* with `display-id-pattern-enforcement:
off` — so the only way to keep readable component names is an explicit
`Type:` on each one.

## Proposed direction

1. Relax `compileDisplayIdPattern` to allow **zero counters** when the
   pattern has a non-empty literal anchor (a prefix), so matches stay
   disambiguable. A bare `{name}` with no literal prefix should still be
   rejected — it would match every ID and collide with all types.
2. Let `classifyEntry`'s existing pattern loop consume counter-less patterns
   unchanged — the single-match / `MSL-T002` ambiguous logic already handles
   prefix collisions.
3. Pair with `display-id-pattern-enforcement: off` (or a `shape` level),
   since there is no counter to enforce — the pattern is for *classification*,
   not numbering.
4. Result: `sw-component: display-id-pattern: "SWC_{name}"` /
   `hw-component: "HWC_{name}"` classify `SWC_LIGHT_CTRL` / `HWC_PIU` with no
   explicit `Type:`.

## Notes

- Preserves the named-entity model — no forced numbering like `SWC_HMI_0001`.
- Related: markspec#582 (component generation tooling) — generated components
  would classify by prefix without emitting `Type:`.
- Downstream: the SEED profile would then drop the explicit `Type:` from
  `sw-component` / `hw-component` entries (sibling to #593, which asks the
  reverse direction — display-ID references resolving to ULIDs).

stasson:
Closes #594.

## Summary

Allow a `display-id-pattern` with **no `{n}` counter** so types whose IDs are
*named, not numbered* — components `SWC_LIGHT_CTRL`, `HWC_PIU` — classify by
prefix without an explicit `Type:` trailer.

A pattern is now one of two kinds, selected by the presence of `{n}`:

- **Numbered** (contains `{n}`) — **unchanged**. Exactly one counter; medial
  `{scope}`-style placeholders stay single-segment; mintable + classifying.
  Every pre-existing pattern contains `{n}`, so **no pattern in the wild
  changes behaviour**.
- **Named** (no `{n}`) — new. Requires a non-empty literal anchor plus a
  trailing named placeholder that captures the **rest of the display ID**
  (`[A-Za-z0-9._/-]`, underscores included), so `SWC_{name}` matches
  `SWC_LIGHT_CTRL`. A bare `{name}` (no anchor) is rejected. Classification
  only — not mintable.

Previously a counter-less pattern crashed `markspec check` with an uncaught
`missing {n} placeholder` the moment classification reached the type.

## Design notes

- The split is gated on the absence of `{n}` (rather than overloading
  placeholder semantics by position), which is why numbered patterns are
  provably unchanged.
- The trailing-name char class includes `_` deliberately — a single-segment
  `[A-Za-z0-9]+` would *not* match the issue's headline `SWC_LIGHT_CTRL`.
- The minting path is already decoupled: `parseDisplayIdPattern` (keyed on
  `{n:Nd}`) returns `undefined` for counter-less patterns, so `next-id` /
  `create` / `insert` and the LSP scaffold skip named types unchanged.
- `classifyEntry`'s single-match / `MSL-T002` ambiguity logic is untouched —
  overlapping named prefixes still emit `MSL-T002`.

## Changes

- `core/validator/pattern.ts` — relax `compileDisplayIdPattern` (numbered vs
  named kinds; rest-of-ID terminal `{name}`; reject bare `{name}` / all-literal).
- Tests — `pattern_test.ts`, `types_test.ts`, new e2e
  `named_pattern_classification_test.ts`.
- Docs — `pattern.ts` grammar header, `language.md §2.2`, annex `§B.3.2`,
  **ADR-025** (refines ADR-009 §5), ADR index in `AGENTS.md`.

## Verification

- `just build` (lint + full test suite + typecheck + compile) — pass
- `deno fmt --check` + `dprint check` — pass

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -421,6 +421,10 @@ docs/
   `HardwareInterface` from `Component` to `Contract` (an interface is a
   specification, not a building block); `Provides`/`Requires` provider/consumer
   links; supersedes the type tree shown in ADR-017
+- `adr-025-counter-less-display-id-pattern.md` — counter-less ("named")
+  `display-id-pattern` so types whose IDs are named, not numbered (components
+  `SWC_LIGHT_CTRL`, `HWC_PIU`) classify by prefix without an explicit `Type:`;
+  refines ADR-009 §5; classification-only (not mintable)
 - `overview.md` — narrative architecture tour
 
 See [docs/architecture/overview.md](docs/architecture/overview.md) for a reading
--- a/docs/architecture/adr-025-counter-less-display-id-pattern.md
+++ b/docs/architecture/adr-025-counter-less-display-id-pattern.md
@@ -1,0 +1,88 @@
+# ADR-025: Counter-less (named) display-id-pattern
+
+## Status
+
+Accepted (2026-06-05). Refines [ADR-009 §5](adr-009-core-profile-boundary.md)
+"Minimum pattern grammar". Addresses issue #594.
+
+## Context
+
+A profile classifies an Authored entry's type two ways: an explicit `Type:`
+trailer, or a `display-id-pattern` match (`core/validator/types.ts`
+`classifyEntry`, step 2). ADR-009 §5 set the minimum pattern grammar as "literal
+prefix + one numeric placeholder `{n}`", and `compileDisplayIdPattern`
+(`core/validator/pattern.ts`) enforced it: a pattern with zero `{n}` counters
+threw `missing {n} placeholder`.
+
+That excludes a whole class of types whose IDs are **named, not numbered** —
+components such as `SWC_LIGHT_CTRL`, `HWC_PIU`. They cannot declare a pattern,
+so `classifyEntry`'s pattern loop never matches them and every such entry must
+hand-write `Type:`. Worse, a profile that _did_ declare a counter-less pattern
+(e.g. `SWC_{name}`) crashed `markspec check` with an uncaught exception the
+moment classification reached that type.
+
+Late-stage inference (`types.ts` steps 5/6/8) only resolves the abstract _core_
+type and emits an `MSL-T021` advisory telling the author to declare `Type:`
+anyway — it does not assign the profile type.
+
+## Decision
+
+**Allow a counter-less `display-id-pattern` for classification.** A pattern is
+now one of two kinds, selected by the presence of the `{n}` counter:
+
+- **Numbered** (contains `{n}`) — unchanged. Exactly one counter; single-
+  segment `{scope}`-style placeholders; mintable and classifying. Because every
+  pre-existing pattern contains `{n}`, this guarantees zero behaviour change to
+  any pattern in the wild.
+- **Named** (no `{n}`) — new. Requires a non-empty literal anchor plus at least
+  one named placeholder. The trailing named placeholder captures the **rest of
+  the display ID** (character class `[A-Za-z0-9._/-]`, underscores included), so
+  `SWC_{name}` classifies `SWC_LIGHT_CTRL`. Classification only — not mintable.
+
+A bare `{name}` (no literal anchor) is rejected: it would match every ID and
+collide with every type. An all-literal template (no placeholder of any kind)
+keeps the historical `missing {n} placeholder` rejection.
+
+```yaml
+profile:
+  types:
+    sw-component:
+      extends: SoftwareComponent
+      display-id-pattern: "SWC_{name}" # ^SWC_(?<name>[A-Za-z0-9._/-]+)$
+      display-id-pattern-enforcement: off
+    hw-component:
+      extends: HardwareComponent
+      display-id-pattern: "HWC_{name}"
+      display-id-pattern-enforcement: off
+```
+
+`classifyEntry`'s single-match / `MSL-T002` ambiguous logic consumes the relaxed
+recognizer unchanged — overlapping named prefixes still emit `MSL-T002`.
+
+## Consequences
+
+- Named component / element types classify by prefix with no explicit `Type:`,
+  preserving the named-entity model (no forced `SWC_HMI_0001` numbering).
+- Pair named patterns with `display-id-pattern-enforcement: off`: there is no
+  counter to enforce, and the pattern exists purely for classification.
+- The minting path is untouched and already decoupled: `parseDisplayIdPattern`
+  (`core/profile/display_id.ts`, keyed on `{n:Nd}`) returns `undefined` for
+  counter-less patterns, so `markspec next-id` / `create` / `insert` and the LSP
+  scaffold skip named types and fall back to author-typed IDs.
+- `markspec.lock` references and other consumers are unaffected — the change is
+  confined to `compileDisplayIdPattern`.
+
+## Alternatives considered
+
+- **Overload `{name}` semantics by position in every pattern** (terminal named
+  placeholder is always rest-of-ID). Rejected: it risks changing behaviour of
+  existing numbered patterns that happen to end in a named placeholder. Gating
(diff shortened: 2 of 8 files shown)


stasson:
## Code review (self-review per AGENTS.md)

### Overview
Relaxes `compileDisplayIdPattern` so a `display-id-pattern` with no `{n}` counter is a valid **named** pattern (literal anchor + trailing `{name}` capturing the rest of the ID), classifying named component/element types by prefix. Numbered patterns (any with `{n}`) are untouched. Ships code + unit + e2e tests + spec/ADR docs.

### Correctness ✅
- **Numbered patterns provably unchanged.** The numbered/named split keys on the presence of `{n}`; every pre-existing pattern contains `{n}`, so none changes behaviour. Verified against `profile_types_test.ts`, `discipline_profile_extension_test.ts`, `discipline_override_freeze_test.ts` — all green.
- **Fixes a real crash.** A counter-less pattern previously threw an uncaught `missing {n} placeholder` mid-`classifyEntry`, taking down `markspec check`. The new e2e test reproduced it (RED) before the fix.
- **Char class is right.** Terminal `{name}` → `[A-Za-z0-9._/-]+` (underscores included) so `SWC_LIGHT_CTRL` matches; a single-segment class would not. Medial `{scope}` stays `[A-Za-z0-9]+`, so `XREQ_{scope}_{n:04d}` is unaffected.
- **Reference entries unaffected** — `classifyEntry` still returns before the pattern loop for non-Authored shapes.
- **Minting stays decoupled** — `parseDisplayIdPattern` (keyed on `{n:Nd}`) already returns `undefined` for counter-less patterns, so `next-id`/`create`/`insert`/LSP-scaffold skip named types unchanged.

### Conventions ✅
- Scope `core`, one Conventional commit, no CHANGELOG entry (batched at release), `Closes #594` in body.
- ADRs are immutable-once-accepted → refinement recorded as **new ADR-025** (not an edit to ADR-009), indexed in `AGENTS.md`.
- `just build` + `deno fmt --check` + `dprint check` all pass.

### Test coverage ✅
- Unit (`pattern_test.ts`): rest-of-ID match incl. underscore/dot/slash, anchored prefix, empty-name rejection, `{scope}` counter-less now valid, bare `{name}` rejected, all-literal still rejected.
- Unit (`types_test.ts`): named classification (no `Type:` → `sw-component`) + overlapping named prefixes → `MSL-T002`.
- E2E: `SWC_LIGHT_CTRL`/`SWC_DSG`/`HWC_PIU` classify clean; non-matching ID still `MSL-T003`.

### Minor / out of scope
- **Duplicate named placeholders** (`SWC_{name}_{name}`) throw a raw `RegExp` "Duplicate capture group name". This is **pre-existing** — numbered patterns hit the same (`X_{scope}_{scope}_{n:04d}`). Same class as the deliberately-deferred "validate patterns at profile-load time → clean `PROFILE-*` diagnostic" item (see ADR-025 *Alternatives considered*). No regression.
- **No per-compile caching** — `classifyEntry`/`checkEnforcement` recompile the pattern per entry. Pre-existing; not worsened.

### Risk: low
Surgical, additive, well-covered; numbered patterns are provably inert.

stasson:
## /review — multi-angle (7 finder angles + verification)

Ran 3 correctness angles (line-by-line, removed-behavior, cross-file tracer) + cleanup/efficiency + altitude over the diff, with empirical probes.

### Verdict: no severe issues introduced — safe to merge

Confirmed by independent finders + probes:
- **Numbered patterns are byte-identical** to pre-change output (`restOfIdNamedIndex` is forced to `-1` when a counter is present) → zero regression for any existing pattern.
- **Rest-of-ID class does not over-match across a literal boundary** — `$`-anchoring + backtracking keep `SWC_{name}_END` correct.
- **`checkEnforcement` / MSL-T004** behaves correctly for named patterns (no false flag, no crash).
- **Real #594 use case is clean** — `SWC_`/`HWC_` named types don't overlap numbered requirement prefixes (`SRS_`, …).

### Minor / pre-existing — filed as debt (not blocking)

- **#596** (bug, pre-existing): `compileDisplayIdPattern` and `parseDisplayIdPattern` disagree — `{n:4d}` is rejected by one and accepted by the other, and the annex documents the rejected form. Unify the two parsers.
- **#597** (debt): a malformed `display-id-pattern` (duplicate named placeholder `SWC_{x}_{x}`, bare `{name}`, …) throws an uncaught error during classification instead of a clean `PROFILE-*` diagnostic at profile-load. This is the load-time-validation item ADR-025 explicitly deferred.
- **#598** (debt): named (counter-less) types are classifiable but not mintable — `next-id`/`create`/`insert` print a misleading "malformed pattern" error and the LSP offers no scaffold. Make those paths named-aware (relates #582).

### Notes (not filed)

- A named pattern's trailing class includes digits, so a named type whose prefix **collides with a sibling numbered type's prefix** (e.g. `REQ-{name}` + `REQ-{n:04d}`) makes both match `REQ-0001` → `MSL-T002`. That is the ambiguity system working as designed (a correct "add explicit Type:" diagnostic, not a wrong result); it does not arise for distinct prefixes. Worth a one-line authoring note.
- `compileDisplayIdPattern` recompiles per entry × per type (no memoization) — pre-existing; the diff adds a small per-call allocation. A `Map<template, RegExp>` memo is a cheap future win on the LSP `validateAll` path. Not filed (pre-existing, sub-perceptible at current scale).

No code changes applied — nothing rose to "severe".

PR #651, opened 2026-07-01
feat(cli): project-scope defaults, gitignore-aware discovery, composite check gate

stasson:
## Smoother CLI: project-scope defaults, gitignore-aware discovery, composite `check`

Makes `markspec check` / `lint` / `fmt` do the obvious whole-project thing when
invoked bare, unifies file discovery behind one gitignore-aware walker, and turns
`check` into the single gate you wire into CI and pre-push. Decision record:
[ADR-027](docs/architecture/adr-027-cli-smoother-defaults.md).

### What changed

- **`core/discovery/`** — a pure-TS, gitignore-aware file walker (with a
  ReDoS-hardened gitignore matcher). Now the single walker for the CLI verbs, the
  LSP indexer, and lockfile collection; replaces three ad-hoc walkers and a
  duplicated extension list.
- **Bare `check` / `lint` / `fmt`** operate on the whole project (gitignore +
  `project.yaml` `exclude:` honored), print a scope header to stderr (suppressed
  by `-q` / `--format json`), and require a discoverable `project.yaml`. Explicit
  paths still give exact scope; directory args expand through the gitignore
  filter; `fmt` is markdown-only.
- **`project.yaml` gains `exclude:`** — gitignore-syntax patterns for the
  whole-project default scope.
- **`check` is a composite gate** when run bare:

  | Gate | Code | Severity |
  |------|------|----------|
  | Structure / attributes / listing | (existing) | as today |
  | Traceability | MSL-L006 | warning |
  | Format drift | **MSL-F010** | error |
  | Lockfile edge-hash drift (offline, when `markspec.lock` present) | **MSL-L212** | error |
  | Prose lint | MSL-Q | advisory (warning) |

  One diagnostics stream, one exit-code computation (0 clean / 1 error / 2
  warnings-only), `--strict` promotes warnings to errors.

### Key design decision

The composite gates (fmt-drift, lockfile, prose) and MSL-L006 are
**project-wide only**. `markspec check <file>` stays a fast *structural* check —
unchanged from today. This surfaced during implementation: `format()` lowercases
modal verbs, so a file-local fmt-drift *error* would hard-fail warning-level
content. Scoping to bare invocation keeps per-file checks smooth and is
consistent with the already-project-wide MSL-L006 / lockfile gates.

`markspec lock` and `check`'s MSL-L212 gate walk the identical file set
(extensions + `exclude`), so a freshly-locked project does not false-positive.

### Known limitation

Bare invocation discovers the project root via `project.yaml` only; a project
activated solely by `.markspec.yaml` reports "no project root found". This is
pre-existing, CLI-wide behavior, now documented (ADR-027, AGENTS.md). Teaching
root-discovery about `.markspec.yaml` is a follow-up touching all commands.

### Follow-ups (non-blocking)

- Document MSL-F010 / MSL-L212 in the governed `docs/spec/language/language.md`
  §8 catalogue (per ADR-012's phased-adoption process).
- Unify the 4th walker in `mcp/project.ts` onto `core/discovery`.
- `markspec init` writes an empty-edge lockfile (could false-positive MSL-L212
  before the first `lock`).

### Verification

`just build` green (3273 passed / 0 failed, lint + typecheck + compile clean);
full e2e suite 564 / 0; `deno fmt --check` and `dprint check` clean. Each of the
13 implementation steps was TDD'd and independently reviewed; a whole-branch
review returned READY.


stasson:
## Code review — medium effort (5 finder angles → 1-vote verify)

Independent review of the composite-gate + discovery diff. Three correctness
findings were independently CONFIRMED against the code (two of them upgrade
"benign/accepted" conclusions from the pre-merge review to real gaps). No merge
blockers for a pre-1.0 tool, but the top three are cheap, worthwhile fixes.

### Correctness

1. **`fmt` and `check` disagree about the corpus — `fmt`'s ref-index ignores `exclude:`** — [`cli/commands/fmt.ts` `collectEntries(projectRoot)`]
   `fmt` builds its ADR-026 canonicalization index via `collectEntries(projectRoot)` with **no** `exclude` arg, while `lock` passes `config.exclude` and `check`'s MSL-L006 is exclude-aware. So `fmt`'s `rewriteToken` (`core/refs/mod.ts:211`) treats a reference to a display ID that exists **only** in an excluded path (e.g. this repo's `skills/*/SKILL.md`) as resolved/canonical, while `check` flags that same reference as unresolved. *Trigger:* a real file references an ID that collides with / only exists in an excluded example entry → `fmt` leaves it, `check` fails it. *Fix:* thread `config.exclude` into `fmt`'s `collectEntries` call (one line; makes fmt/lock/check agree).

2. **MSL-F010 misses ref-canonicalization drift → `check` can falsely report "formatted-clean"** — [`cli/commands/check.ts:114`]
   The gate calls only `format(content).changed`; `fmt` additionally runs `canonicalizeRefs`/`buildRefIndex` and sets `changed` on ref-heal drift even when `format().changed === false`. So bare `check` exits 0 on a file `fmt --check` would still flag (a non-canonical but formatter-clean display-ID reference), giving CI a false all-clean. False-negative only (no false positives). *Fix:* run `canonicalizeRefs` in the gate too (check already has the entries + project root), or state the format-only scope prominently in the docs.

3. **gitignore `baseDir` prefix escape is incomplete (`* ? [ ]` unescaped)** — [`core/discovery/gitignore.ts:117`]
   `parseGitignore` escapes the nested-`.gitignore` baseDir with `/[.+^${}()|\\]/g`, which omits `*`, `?`, `[`, `]`. `parseGitignore("*.md", "pkg[v2]")` builds `^pkg[v2]/[^/]*\.md$` — `[v2]` becomes a character class, so the nested rule silently mis-scopes to `pkgv/…`/`pkg2/…` instead of the literal `pkg[v2]/…`. *Trigger:* a directory named with `[`/`]`/`*`/`?` (legal on POSIX) containing its own `.gitignore`. *Fix:* escape baseDir with a complete metacharacter set (and consider sharing one escaper — the set is duplicated from `REGEX_SPECIALS`).

### Behavior change / migration

4. **Lost unconditional build-output skips** — [`lsp/server.ts` + `cli/commands/lock.ts`, deleted `SKIP_DIRS`]
   The old walkers unconditionally skipped `node_modules`/`target`/`dist`/`build`; the new discovery skips only hidden dirs + `.gitignore` + `exclude:`. A consumer project that doesn't list these in `.gitignore` now recurses into build artifacts (duplicate-ID noise, slow LSP indexing, spurious entries in the MSL-L212 edge hash). This repo's own `.gitignore` masks it. *Fix:* keep a small built-in skip set for common build outputs, or document the `.gitignore` dependency.

5. **MSL-L212 hard-errors on upgrade for source-entry projects** — [`cli/commands/check.ts` L212 gate + `cli/commands/lock.ts` widened walk]
   `lock`'s walk widened markdown-only → markdown+source. A pre-existing project with trace attributes in source doc comments (the demo-aeb-* V-model repos) that upgrades and runs bare `check` in CI **before** re-running `lock` gets a hard MSL-L212 error even though requirements didn't change. Documented as an accepted consequence in ADR-027; consider softening the first-run experience (warn, or a release note calling out "re-run `markspec lock`").

### Minor

6. **MSL-L212 emitted with `location: undefined`** — [`cli/commands/check.ts` ~L148] — unlike MSL-F010 (which sets a location), so `--format json` consumers grouping by `location.file` (`lsp/diagnostics.ts` `groupDiagnosticsByFile`) drop L212 from per-file views. Attach the `markspec.lock` path.

7. **Redundant `project.yaml` load on the CLI startup path** — [`cli/helpers.ts` `resolveScope` + `cli/commands/check.ts`] — bare `check` discovers the root and loads+parses `project.yaml` 2–3× (resolveScope for `exclude`, then check again for `captionConventions`). Thread the loaded config through instead of re-loading.

8. **Composite gates are inline CLI glue (altitude)** — [`cli/commands/check.ts:108-176`] — the three `if (scope.projectWide)` gate blocks hand-build diagnostics in the command rather than being first-class stages in `core/validator`'s pipeline (which already threads a `projectWide` flag). A 4th gate means copy-pasting the pattern again. Consider consolidating gate composition into core.

---
*Not flagged: `check <file>` / explicit-dir ignoring `exclude:` (documented by-design in ADR-027); `lint <file>` requiring `project.yaml` (pre-existing; docs correctly don't claim lint works file-locally). `core/discovery` is properly Node-compatible with no `Deno.*` and correct library-boundary usage.*

🤖 medium-effort review via [Claude Code](https://claude.com/claude-code)

stasson:
Pushed three focused fixes for the confirmed review findings (branch updated to `ee9b780`; pre-push `just check` green — 3275 passed / 0 failed). Each is TDD'd (failing test first, then the fix):

- **#1 `fmt` ignored `project.yaml` `exclude:` when building its ref index** — `59a49c7` `fix(cli): fmt honors project.yaml exclude when building ref index`. `fmt` now loads the config and passes `config.exclude` into `collectEntries`, so its ADR-026 canonicalisation corpus matches `check` (MSL-L006) and `lock`. An entry in an excluded path (e.g. `skills/`) no longer resolves a trace reference. New e2e: `tests/e2e/fmt_exclude_ref_index_test.ts`.

- **#3 gitignore baseDir regex escape incomplete** — `1b61e73` `fix(core): complete gitignore baseDir regex escape (* ? [ ])`. `parseGitignore` now escapes the nested-`.gitignore` baseDir prefix with the full metacharacter set (`* ? [ ]` added), so a dir literally named e.g. `pkg[v2]` no longer compiles `[v2]` as a character class. The pattern-body helpers stay narrow by design (comment added). New unit case in `gitignore_test.ts`.

- **#6 MSL-L212 has no location** — `ee9b780` `fix(cli): attach markspec.lock location to MSL-L212`. The lockfile-drift diagnostic now carries `location: { file: <projectRoot>/markspec.lock, line: 1, column: 1 }` (like MSL-F010), so `--format json` consumers that group by `location.file` no longer drop it. Extended the MSL-L212 drift test in `check_project_test.ts` to assert the JSON location.

Left as follow-ups (out of scope here): #2 (MSL-F010 vs `canonicalizeRefs` drift — design call on running canon in `check`), #4 (built-in `node_modules`/`target`/`dist`/`build` skips — design call on a built-in skip set vs documenting the `.gitignore` dependency), #5 (MSL-L212 upgrade hard-error softening), and #8 (lifting the inline gate blocks into core pipeline stages).

stasson:
Merged via `5d9cd39`. The four deferred review findings are now tracked:

- #656 — [design] MSL-F010 drift gate doesn't run `canonicalizeRefs` (check/fmt disagree on "formatted") — `debt`/`validate`/K1/M
- #657 — [design] discovery no longer skips `node_modules`/`target`/`dist`/`build` by default — `debt`/`core`/K1/M
- #658 — soften MSL-L212 hard error on first-run/upgrade — `debt`/`validate`/K2/S
- #659 — lift inline `check` gate blocks into core pipeline stages — `debt`/`core`/K2/L

(Also: the merge-blocking windows-latest failures were pre-existing test-only path-portability bugs, fixed in `474eb1d`.)

Issue #669, opened 2026-07-02
fmt/lsp: dprint-markdown loader hardening (frozen configs, clean load-failure error, LSP MSL-F011 visibility)


stasson:
## Summary

Three small hardening items deferred from the ADR-029 whole-document formatting reviews (none blocks correctness; all improve robustness/UX):

1. **Freeze the exported style constants.** `MARKSPEC_MARKDOWN_GLOBAL_CONFIG` / `MARKSPEC_MARKDOWN_PLUGIN_CONFIG` (`core/formatter/dprint.ts`) are mutable objects exported through the public barrel. `Object.freeze` protects the "one canonical form" invariant against accidental consumer mutation.
2. **clig.dev-clean error on WASM load failure.** `await loadMarkdownFormatter()` in `cli/commands/fmt.ts` and `cli/commands/check.ts` is unguarded — a corrupt/missing plugin surfaces as a raw stack trace instead of a clean one-line error. It fails safe (before any write), but the message should be humane. (The rejected-promise cache poisoning itself was already fixed on the ADR-029 branch.)
3. **LSP MSL-F011 visibility.** `onDocumentFormatting` discards `result.diagnostics`, so a body-polish fallback (MSL-F011 info) is silent in the editor while the CLI prints it. Consider routing format-time diagnostics through the publish-diagnostics channel or a window/logMessage.

## Origin

Final whole-branch review of the ADR-029 PR (whole-document Markdown formatting, #649).

Issue #668, opened 2026-07-02
fmt: ref-canonicalisation rewrites trace values inside fenced code examples


stasson:
## Summary

The `markspec fmt` write path's display-ID/trace-reference canonicalisation (ADR-026, `canonicalizeRefs` applied in `cli/commands/fmt.ts` after `format()`) rewrites ULID/display-ID values **inside fenced ```markdown code blocks** — illustrative examples that the parser/validator correctly treat as inert.

## Impact

Running bare `markspec fmt` on this repo dirties 3 spec/ADR docs on every run (fenced examples get their example IDs "healed" against the real project ref index). Discovered while dogfooding ADR-029's whole-document formatting (PR for #649); the ADR-029 passes are unaffected (both the prose pass and the body polish skip fenced regions by construction — this is exclusively the ADR-026 canonicalisation path).

`check`'s MSL-F010 gate is NOT affected (it calls `format()` without a ref index), so CI stays green — only write-path runs churn the files.

## Expected

Ref canonicalisation must honor the same verbatim-region rules as every other fmt pass: content inside fenced code blocks (and indented code) is never rewritten.

## Where

`core/refs/` + its application in `cli/commands/fmt.ts` (ADR-026, shipped in PR #604). During the ADR-029 work the three affected files were reverted by hand each run; see the ADR-029 PR discussion for the file list.

PR #670, opened 2026-07-02
feat(core): whole-document Markdown formatting in markspec fmt (ADR-029)

### LINKED ISSUE #649: fmt: markspec fmt does not normalize Markdown tables — decide dprint vs. native
## Summary

`markspec fmt` does not normalize/align Markdown tables inside entry bodies.
Misaligned or ragged tables are left completely untouched. We should decide the
intended division of labor: should `markspec fmt` align tables natively, or is
table formatting deliberately delegated to `dprint`?

## Current behavior

```markdown
- [STK_0006] Misaligned table

  Intro prose.

  | Mode | Longer heading |
  |--|--|
  | Fast | x |
  | Safe | a much longer cell value |

      Id: 01JADYKACKQKGVGHT9K7Y6PBPA
```

```
$ markspec fmt misaligned.md
0 file(s) formatted, 1 unchanged (1 total)
```

The file is unchanged — the ragged table is not aligned, the header separator
row (`|--|--|`) is not normalized.

## Why this matters

- Authors reasonably expect `markspec fmt` to be *the* formatter for MarkSpec
  documents (it already stamps ULIDs, normalizes trailer indentation, adds
  trailing backslashes). Tables being silently skipped is surprising.
- The repo's stated formatting split is: `deno fmt` for TypeScript, **`dprint`
  for Markdown/JSON/YAML/TOML** (both are CI gates). `dprint-plugin-markdown`
  *does* align tables. So today, tables get formatted only when a file passes
  through `dprint` — not through `markspec fmt`.

## Decision to make

1. **Delegate to dprint (document it).** Keep `markspec fmt` focused on
   entry-block semantics (Id/trailer/backslash) and rely on `dprint` for table
   alignment. If so: document this explicitly (guide + the entry-authoring
   skills), and make sure the `markspec fmt` → `dprint` ordering is safe.
   - ⚠️ Interaction to verify: `dprint` operates on the whole file and is not
     entry-trailer-aware. Confirm it never reflows/wraps trailer attribute
     values or the 6-space indented trailer code block. Note that
     `docs/examples/` is already excluded from `dprint` for exactly this
     "must not line-wrap long attribute values" reason.

2. **Native table normalization in `markspec fmt`.** Have the entry-aware
   formatter align table columns itself (it already parses tables into
   `bodyAst[].kind === "table"` with `header`/`rows`/`raw`), so a MarkSpec
   project needs only one formatter and tables are aligned even without a
   `dprint` pass.

## Note

This is adjacent to the parser bug where colons in table cells are misread as
trailer attributes — a table-aware formatter and a table-aware parser share the
same "recognize table block content" need.

## Environment

markspec 0.8.0 (core-schema 1), Deno 2.8.3, macOS 15 (Darwin 24.6.0)

stasson:
Closes #649.

## Summary

`markspec fmt` now formats the **entire Markdown document** — prose chapters, headings, tables, lists, and entry bodies — not just entry-block mechanics. Implements **ADR-029** (included in this PR): default-on, zero config, Markdown files only, via the embedded dprint-markdown WASM plugin (`@dprint/formatter@0.5.1` + `@dprint/markdown@0.20.0`, exact pins, lazy-loaded only on formatting code paths).

## How it works

- **Segmentation (approach B):** entry title lines and trailers never reach dprint. Prose segments between entries are formatted whole; entry bodies are polished dedented at `lineWidth − indent` (so re-indented lines stay ≤ 80 and agree byte-for-byte with an external dprint view — verified against this repo).
- **Safety gates:** every dprint rewrite must pass a new CommonMark-semantic equivalence comparator (mdast-based, ASCII-whitespace-collapse only — NBSP is content); rejected rewrites fall back with an advisory `MSL-F011` (info). ADR-015's strict `astEquivalent` / `MSL-F900` guard is untouched.
- **Idempotency proven:** `format(format(x)) === format(x)` over the fixture corpus AND raw in-memory ragged documents, with the real plugin (`MSL-F010` depends on this).
- **All three surfaces wired identically:** CLI `fmt`/`fmt --check`, `check`'s project-wide MSL-F010 drift gate, and LSP format-on-save produce byte-identical output.
- **Fixed style:** lineWidth 80 (soft target — table rows, links/URLs, inline code may exceed it, never split), `textWrap: always`, `_emphasis_`, `**strong**`, `-` lists; line endings preserved per file. `<!-- dprint-ignore -->` works (ranges must not span an entry block).
- **Scope guard:** the Markdown pass applies to `.md` files only — explicit source-file args and LSP formatting requests on source files are structural no-ops (regression-tested on both paths).

## Breaking / release-notes call-outs

- Existing projects need a **one-time `markspec fmt`** to converge; `check` (MSL-F010) now flags prose drift too (pre-1.0 no-migration policy).
- Semantic line breaks (one-sentence-per-line authoring) do not survive `textWrap: always`.
- Deliberately-unformatted files use `project.yaml` `exclude:` (this repo adds `docs/examples/`, `docs/templates/`, `CHANGELOG.md`, `docs/product/ast-fidelity-matrix.md`).
- Future `@dprint/markdown` version bumps change the canonical form — pins are exact; bumps are release-notes events.

## Dogfooding results

Bare `markspec fmt` on this repo converges with the external dprint CLI (zero ping-pong; `dprint check` clean). The run surfaced one pre-existing bug now filed as #668 (ADR-026 ref-canonicalisation rewrites fenced examples — unrelated to this PR's passes). Deferred hardening items are #669.

## Verification

- Full `just build` + `deno fmt --check` + `dprint check` green; 2754+ unit tests, e2e suites green.
- Compiled-binary smoke: the WASM loads and formats inside `dist/markspec` (Route A, npm-embedded).
- Every task was implemented via TDD and independently reviewed (spec + quality) with fix rounds; a final whole-branch review (including an adversarial pass on the safety gates) concluded ready-to-merge after closing a source-file scope hole with regression tests on both the CLI and LSP paths.

Spec + implementation plan are archived under `docs/archive/` (gardened; `docs/wip/` empty).


stasson:
## /review findings (precision pass: 8 finder angles → verified 1-vote each)

None of these blocks the merge — the safety ladder, idempotency proof, and scope guards shipped in this PR held up under adversarial verification. Findings ranked most-severe first; 1–2 are worth a fast-follow (or pre-merge fixup), the rest are hardening/cleanup.

**1. `core/formatter/mod.ts:314` + `core/formatter/prose.ts:63` — unguarded `proseFormat()` calls: a dprint WASM formatting error crashes the whole run instead of falling back.** *(CONFIRMED, empirically reproduced)* `@dprint/formatter`'s `formatText` throws on plugin errors (`@throws If there is an error formatting`). There is no try/catch between these call sites and `format()`'s boundary, and none in the `fmt`/`check` per-file loops: one pathological file aborts the remaining files in `markspec fmt` and kills the entire `markspec check` invocation. (LSP degrades gracefully — vscode-jsonrpc converts the rejection to an InternalError response.) **Fix:** wrap both call sites and treat a throw exactly like a semantic-gate rejection — keep the original text, emit the MSL-F011-style advisory. This completes the "never make a file worse / never block on the new pass" ladder for the last uncovered failure mode.

**2. `core/formatter/mod.ts:467` — the "Markdown files only" invariant lives only in call-site guards; `format()` itself never consults `options.file`.** *(CONFIRMED)* Three independently-written guards (fmt.ts `MARKDOWN_EXTENSIONS`+`extname`, check.ts `.endsWith(".md")`, lsp `isSourceFile`) protect today's callers, but a future caller that wires `formatMarkdownProse` without its own guard silently corrupts source files — and `core/formatter/source.ts:formatSource` already forwards options into `format()` with source-file labels. The semantic gate cannot catch it (line-joining is wrap-equivalent as Markdown). **Fix:** gate inside `format()` (skip the markdown pass unless `options.file` has a Markdown extension); the call-site checks become defense-in-depth. All existing callers/tests already pass `.md`-suffixed file labels, so this breaks nothing.

**3. `cli/commands/check.ts:81` — case-sensitive `.endsWith(".md")` breaks fmt/check parity for oddly-cased files.** *(CONFIRMED, empirically reproduced; pre-existing, widened by this PR)* Discovery and fmt both lowercase extensions, so `README.MD` is discovered and reformatted by `fmt` — but it never enters `mdContents`, so MSL-F010 never runs on it: `fmt --check` exits 1 while `check` stays green on the same file. Predates this branch (introduced with the F010 gate), but the PR widens the hidden-drift class from Id-stamping to whole-document prose. **Fix:** `MARKDOWN_EXTENSIONS.has(extname(filePath).toLowerCase())` at check.ts:81.

**4. `cli/helpers.ts` (`resolveScope`) — explicitly-named files bypass `project.yaml` `exclude:`, and the wrap pass makes that destructive for the first time.** *(PLAUSIBLE, repro'd on a copy of `docs/examples/entry-rendering.md`: ~15 paragraphs re-wrapped, deliberate one-clause-per-line EARS examples collapsed; trailers/fences untouched.)* "Explicit args = exact scope" is the documented CLI convention and `docs/guide/cli.md` states the caveat — but ADR-029's Consequences bullet ("Deliberately-unformatted files use `project.yaml` `exclude:`") is unqualified, and `fmt` gives no runtime notice. **Fix:** print an info line to stderr when an explicitly-named path matches `exclude:`, and add the caveat to ADR-029.

**5. `core/formatter/md_equiv.ts:27` — private remark pipeline duplicates the exported shared one.** *(CONFIRMED)* `core/parser/remark.ts` already exports `processor = unified().use(remarkParse).use(remarkGfm)` — the identical plugin set — and formatter→parser imports are established convention. If the shared pipeline ever gains a plugin, the equivalence gate silently diverges from the parser. **Fix:** `import { processor } from "../parser/remark.ts";`.

**6. `core/formatter/prose.ts:55-79` / `core/formatter/mod.ts:304-330` — the polish→gate→fallback sequence and the MSL-F011 wording exist in two hand-written copies.** *(PLAUSIBLE — cleanup)* Verification concluded a shared helper would need too many parameters to be a win (input shapes and fallback targets genuinely differ), so the actionable form is smaller: cross-referencing comments at both sites + aligning the two MSL-F011 message shapes, so a future tolerance change or wording fix can't land in only one copy.

**Verified-and-dropped** (for the record): the "triple parse per changed file" is the correct simple choice (body splices invalidate pre-splice entry positions); LSP first-format WASM load measured at ~12 ms (negligible); the unguarded-loader stack trace and mutable exported config objects are already tracked in #669.

Issue #675, opened 2026-07-02
Prose lint never runs on profile-typed entries (isProseScope is profile-blind)


stasson:
Pre-existing bug discovered during the ADR-030 feature review (not introduced by that branch; `core/lint/runner.ts` last touched around `c4d0ba6`):

`core/lint/runner.ts:60` calls `resolvedCoreType(entry)` WITHOUT the profile argument, but `resolvedCoreType(entry, profile?)` (`core/validator/type_resolution.ts:187`) can only map profile-declared types (e.g. `Type: platform-component`, or a profile display-id-pattern classification) to their core ancestors when the profile is passed.

Consequence: entries classified purely via a profile type never enter prose scope — `markspec lint` and `check`'s advisory prose gate silently skip most profile-typed requirement entries project-wide. Only entries with explicit core `Type:` values, source introspection, or core-recognizable display-ID prefixes (e.g. `REQ_`) get prose-linted.

Fix direction: thread the active `EffectiveProfile` into `runLint`/`isProseScope` so `resolvedCoreType` can walk the profile type's `extends:` chain.

Repro: a profile declaring a Requirement-shaped type with its own display-id pattern, and an entry classified by that pattern whose body contains "The bus shall be exposed by the service." — `markspec lint` reports nothing; the same sentence under a core-recognizable ID fires MSL-Q300 (passive voice).

Issue #674, opened 2026-07-02
Delivered-documents (ADR-029) follow-up bundle: tests, consolidation, polish


stasson:
Follow-up debt from the ADR-030 profile-delivered-documents feature branch (`feat/profile-delivers`), triaged by the final whole-branch review as non-blocking:

**Tests**

- MSL-R005/I007/I008 suppression paths in `attributeCorpusDiagnostics` untested (only R006 exercised); document the message-quote-format coupling with the validator in a code comment.
- `detectCorpusCollisions` Id-(ULID)-collision branch has no dedicated unit test.
- MCP cache-hit path returning merged corpus diagnostics covered by inspection only (a two-call `getCompiled()` assertion would pin it).

**Consolidation**

- `profileId@profileVersion` label idiom at 5 sites (`core/validator/corpus.ts`, `core/profile/delivered.ts`, `cli/commands/show.ts`, `core/reporter/mod.ts`, `mcp/resources/entry.ts`) — add a shared `formatEntryOrigin(origin: EntryOrigin)` next to the model type; `corpusOriginLabel` remains the `DeliveredDocument` twin.
- `buildCorpusIndex` built at 3 CLI sites; corpus-load block duplicated between `cli/helpers.ts` and `cli/commands/check.ts` — have `compileProject` return its corpus index.

**Polish**

- LSP: one-time `sendDiagnostics(uri, [])` when a path first enters `corpusFilePaths` (closes the session-restored-corpus-tab stale-diagnostics race); LSP surfaces generic MSL-R006 instead of MSL-R014 for corpus collisions (`validateAll` does not run `detectCorpusCollisions`).
- `language.md` §8.2 MSL-R014 row says "Rename the project entry" — imprecise for the inter-tier corpus-to-corpus case.
- Annex B §B.15 "Load-time diagnostics" intro sentence describes only -001/-002 though the table includes parse-time -003/-004.
- MCP delivered resources hardcode `mimeType: text/markdown` for any file type; doctor's delivered-health section can only ever show the docs-only-missing warning (error-severity issues exit via `compileProject` first).

stasson:
Adding the four findings from the post-PR `/review` on #676 that were left for a fast-follow (the review comment is on the merged PR). All confirmed against the shipped code:

[...]

PR #676, opened 2026-07-02
feat(core): profile-delivered documents — corpus + docs shipping via profiles (ADR-030)

stasson:
## Profile-delivered documents (ADR-030)

A profile package can now deliver document files to consuming projects. Each `profile.delivers:` item is either a **traceable corpus** (`corpus: true` — its entries join the consumer's traceability graph as read-only, provenance-marked citizens) or **documentation-only** (surfaced as MCP resources). This turns profiles from vocabulary packages into content packages: compliance profiles can ship a standard's objectives as satisfiable entries; platform teams can ship a reference architecture that downstream projects `Allocated-to:`/`Realizes:` against; orgs can share requirement baselines.

### What shipped

- **Manifest**: `profile.delivers:` section (path/corpus/description) with path-escape validation (`PROFILE-DELIVERS-003/-004`); JSON schema + Annex B §B.15 updated.
- **Chain merge**: `EffectiveProfile.delivers` — additive union across the `extends:` chain, parent-first, each tier's paths resolved against its own package dir; children cannot remove a parent's delivery.
- **Core loader**: `loadDeliveredCorpus` parses corpus files once and stamps `Entry.origin` (`{ kind: "profile", profileId, profileVersion }`); missing corpus file = error (`PROFILE-DELIVERS-001`), missing docs-only file = warning (`-002`).
- **One loader, three callers**: compiler (`CompileOptions.corpusEntries`), LSP (`WorkspaceIndex` seeded corpus-first before the project walk; re-seeded on profile reload; cold-cache soft-fail), MCP (compiled context + diagnostic parity with `check`).
- **Validator**: new `MSL-R014` — project↔corpus and inter-tier corpus↔corpus ID collisions are hard errors naming the owning profile; corpus-located findings are downgraded to warnings attributed `delivered by <id>@<version>:` so consumer builds never go red on upstream content. Prose lint and `--strict` promotion are corpus-blind.
- **Read-only by construction**: `fmt`, rename (LSP guard + edit-loop guard), document formatting, and document-sync events never touch corpus files; the `MSL-L212` lock gate ignores corpus edges (lockfile integration is an explicit non-feature).
- **CLI surfaces**: corpus locations render as `<profile-id>@<version>:<relative-path>:<line>` (never raw cache paths, incl. `show`'s Source line); `show` prints `Origin:`; `profile show` gains a Delivered documents block (with missing-file indicators); `doctor` reports delivered-file health; traceability matrix + `export json|yaml|csv` carry an `origin` column/field.
- **MCP**: `markspec://delivered/{profileId}/{path}` resources (traversal-safe reads), Delivered documents section in `markspec://profile`, origin line in entry views.
- **LSP**: corpus IDs in completions from first index (badged `from <id>@<version>`), hover/go-to-def work, rename refuses corpus targets.
- **Docs**: ADR-030 (decisions D1–D8 incl. the no-`CORE_SCHEMA_VERSION`-bump correction and the local-profile `exclude:` requirement), `language.md` §8.2, guide (`profiles.md`, `cli.md`) + new recipe *Shipping a reference architecture in a profile*. Spec + plan archived under `docs/archive/`.

### Known operational requirement

Local-path profiles (`./profile`) must be excluded from project discovery (`project.yaml` `exclude:`), otherwise the corpus file is parsed twice and self-collides (MSL-R014). Documented in the recipe and ADR-030 D7; git/npm profiles live under `.markspec/cache/` which discovery already skips.

### Testing

TDD throughout (unit + blackbox e2e). Key proofs: a differential e2e pair (`check` with `delivers:` resolves the corpus `Satisfies:` target; the same fixture without it emits MSL-L006), RED-verified regression tests for the lock gate, `--strict` exemption, prose-lint filter, collision codes, and `profile show` health rows; byte-determinism e2e for `export json`. `just build` green; `deno fmt --check` + `dprint check` clean; compiled binary smoke-tested.

Every task passed an independent spec+quality review; a final whole-branch review (verdict: ready to merge) drove one last fix wave — its non-blocking triage is filed as #674, and it uncovered pre-existing #675.

### Follow-ups

- #674 — test/consolidation/polish bundle (non-blocking triage from the final review)
- #675 — pre-existing: prose lint never runs on profile-typed entries (`isProseScope` profile-blind)

> **Note (renumber):** main took ADR-029 for whole-document markdown formatting (PR #670) while this branch was in flight; this feature is now **ADR-030** everywhere (branch merge commit 2039896). The review comment above predates the renumber — its ADR-029 mentions refer to this feature.


stasson:
## /review — PR #676 (profile-delivered documents, ADR-029)

Post-PR review pass: 8 independent finder angles over the full diff (66 files, +5469/−56), 30 candidates deduped to 9, each verified by an independent agent against the checked-out head (2 refuted, 7 survived). Findings ranked most-severe first.

### Findings

**1. `check.ts` prose-lint pre-filter breaks the MSL-Q500 identifier index's additive contract** — `packages/markspec/cli/commands/check.ts:218`
`runLint({ entries: allEntries.filter((e) => !e.origin) })` removes corpus entries from the very array `buildIdentifierIndex` scans, violating the documented ADR-021 invariant on `core/lint/rules/xref.ts` ("more entries → more silenced phrases, never more diagnostics"). A `$Identifier` defined only in a delivered corpus document no longer silences Q500, so a project entry that legitimately references `$SomeCorpusEntity` now trips a spurious `xref-glossary-undefined` warning. Reachable today (the e2e fixture's `REQ_`-prefixed entries prove project entries hit the Q500 scan). Fix shape: lint the full entry set and post-filter the *output* diagnostics by corpus location (the pattern `attributeCorpusDiagnostics` already uses), instead of pre-filtering the input.

**2. Standalone `markspec lint` (and `--strict`) still runs prose analysis on corpus entries** — `packages/markspec/cli/commands/lint.ts:33`
`lint.ts` passes `compileProject()`'s merged entry set (which now includes origin-stamped corpus entries) into `runLint` with no filter — only `check`'s embedded gate was made corpus-blind. ADR-029 D4 states "prose lint does not run on them at all." Reachable today for corpus entries with core-recognizable prefixes (`REQ_`, `TST_`, …) or explicit core `Type:`: unfixable upstream findings appear, the `score` roll-up skews, and `lint --strict` exits 1 on content the consumer cannot edit. (Interacts with #675: once `isProseScope` becomes profile-aware, exposure widens to all profile-typed corpora.) Same fix shape as finding 1 covers both.

**3. LSP leaks raw profile-cache paths into the editor for corpus collisions** — `packages/markspec/lsp/workspace.ts:257` (validateAll)
`validateAll` never runs `detectCorpusCollisions`, so typing a corpus-owned display ID in a project file publishes core MSL-R006 whose message embeds the corpus entry's absolute path: `duplicate display ID 'PLT_0001' (also at /Users/<user>/.markspec/cache/<sha>/reference/platform.md:3)`. The diagnostic is located in the *project* file, so the `corpusFilePaths` publish guard doesn't filter it, and `toLspDiagnostic` copies the message verbatim — contradicting ADR-029's "human-facing diagnostics never show raw cache paths," and diverging from `check`'s MSL-R014 for identical state. (The R006-vs-R014 half is noted in #674; the cache-path leak is the sharper, new part.)

**4. `export`/`compile --format json` silently omit corpus-load warnings** — `packages/markspec/cli/helpers.ts:206`
`compileProject` prints `corpus.diagnostics` to stderr (and exits on errors) but never merges them into the returned `CompileResult.diagnostics`, so serialized artifacts omit warning-severity issues like PROFILE-DELIVERS-002 — while MCP (`mcp/project.ts` merges with a "CLI check parity" comment) and `check` both report them. A CI consumer parsing `export json` believes the corpus is healthy when other surfaces flag it.

**5. `doctor`: double corpus load, duplicated stderr diagnostic, and exit-2 conflation** — `packages/markspec/cli/commands/doctor.ts:78,182`
(a) The corpus is read+parsed twice per invocation (`compileProject([])` loads it, then doctor re-calls `loadDeliveredCorpus`); (b) live-reproduced: a PROFILE-DELIVERS-002 warning prints twice in two formats (once via compileProject's loop, once in the "Delivered documents" block); (c) `if (belowFloor || corpusIssues.length > 0) Deno.exit(2)` conflates delivered-doc health with toolchain skew while the comment above it still says exit 2 means "CI can gate on toolchain skew" — a `doctor`-exit-2-means-upgrade script now mis-fires on a missing docs-only file. Fix shape: have `compileProject` return the corpus diagnostics it already computed, and either give corpus issues their own exit path or update the comment/docs.

**6. `report traceability` Origin column inserted mid-order breaks positional CSV consumers** — `packages/markspec/core/reporter/mod.ts:139`
Header changed from `ID,Title,Type,Satisfies,Satisfied-by` to `ID,Title,Type,Origin,Satisfies,Satisfied-by`, shifting Satisfies from field 4 to 5 — while the same PR's `export csv` deliberately *appended* origin last for exactly this positional-compat reason. Pre-1.0 posture permits breaks, but this one is avoidable at zero cost (append last, matching the export convention); the e2e only asserts a header prefix so nothing pins the choice.

**7. Missing doc comments on exported corpus functions** — `packages/markspec/core/validator/corpus.ts:47,126`
`detectCorpusCollisions` and `attributeCorpusDiagnostics` are exported through `core/mod.ts` (the library boundary) with no per-item `/** */` doc comments — AGENTS.md: "doc comments on all public API items." (Verified by inspection; the module header describes them but the file's own siblings carry per-function comments.)

### Refuted during verification (for the record)

- *MCP reloads the corpus on every recompile*: recompiles are already unconditional full rebuilds of the (larger) project set; corpus reload is proportionally trivial and is the only mechanism that picks up local-profile corpus edits.
- *`parseDeliversSection` path check diverges from `ABSOLUTE_RE`*: empirically equivalent on every tested input after the manifest's backslash-normalization; the manifest gate is strictly ≥ as strict and additionally blocks `..` traversal, and importing from `validator/` into `profile/` would create a dependency cycle.

### Suggested disposition

Findings 1–4 are worth fixing before merge (1+2 share one fix; 3 and 4 are each small). Findings 5–7 can fold into the #674 follow-up bundle if preferred.

PR #677, opened 2026-07-02
fix(core): harden and scope the ADR-029 markdown pass (#670 review follow-ups)

stasson:
## Summary

Three fast-follow fixes from the `/review` of #670 (ADR-029 whole-document Markdown formatting). All harden or correctly scope the markdown pass; none changes the feature's happy path.

## Fixes

**1. Guard the two dprint call sites against WASM formatting errors (was: crash).** `@dprint/formatter`'s `formatText` throws on plugin errors, and neither call site (`emitBodyViaAst` body polish, `formatProseSegments` prose pass) was wrapped — one pathological file would abort a whole `markspec fmt` run and crash `markspec check`. Both now degrade through the existing fallback ladder: keep the original/canonical text, emit the advisory `MSL-F012` (info), continue. The prose path routes a throw through the identical `fallbackStarts` channel a gate-rejection already uses.

**2. Move the "Markdown files only" invariant inside `format()`.** The scope guard lived only at the three call sites (fmt.ts, check.ts, lsp/server.ts); a future caller wiring `formatMarkdownProse` without its own guard would silently reflow a source file's doc comment as Markdown — and the CommonMark-semantic gate cannot catch that class (line-joining is wrap-equivalent as Markdown). `format()` now self-disables the pass unless `options.file` has a Markdown extension (single source of truth via `MARKDOWN_EXTENSIONS`). The call-site guards remain as defense-in-depth. Notably `core/formatter/source.ts` forwards options into `format()` with source-file labels — now safe.

**3. Case-insensitive `.md` match in `check`'s MSL-F010 gate.** `check.ts` used case-sensitive `filePath.endsWith(".md")`, so `README.MD` was reformatted by `fmt` (which lowercases) but never entered the drift gate — `fmt` reported drift while `check` stayed green. Now `MARKDOWN_EXTENSIONS.has(extname(filePath).toLowerCase())`, matching `fmt` and discovery. Pre-existing hole (predates #670), widened by this feature to whole-document prose drift; the RED run confirmed the discovery walk itself already lowercases, so the fix is isolated to `check.ts`.

## Tests

5 new tests, all confirmed RED before implementation:
- `format()` does NOT call the markdown callback for a `.rs` file even when it's wired (+ a `.md` control that proves it still runs).
- A throwing formatter degrades to fallback (entry-body path emits `MSL-F012`; prose-segment path keeps the original) — neither crashes `format()`.
- `check` on `docs/OVERVIEW.MD` (uppercase) fires `MSL-F010` — case parity with `fmt`.

Full `just check` green (pre-push); `deno fmt --check` + `dprint check` clean.

Follow-ups #668 (fenced-example ref-canonicalisation) and #669 (loader hardening: frozen configs, clean load-failure error, LSP MSL-F012 visibility) remain open.

### DIFF
--- a/.followup677.diff
+++ b/.followup677.diff
@@ -1,0 +1,221 @@
+diff --git a/packages/markspec/cli/commands/check.ts b/packages/markspec/cli/commands/check.ts
+index a9cd2fc7..e6a3e123 100644
+--- a/packages/markspec/cli/commands/check.ts
++++ b/packages/markspec/cli/commands/check.ts
+@@ -4,8 +4,9 @@
+  * `markspec check` — check broken refs, missing Ids, duplicates.
+  */
+ 
++import { extname, join } from "@std/path";
+ import { Command } from "@cliffy/command";
+-import { ConfigError } from "../../core/mod.ts";
++import { ConfigError, MARKDOWN_EXTENSIONS } from "../../core/mod.ts";
+ import type { CaptionConventions, Diagnostic } from "../../core/mod.ts";
+ import { loadActiveProfile, readFile, resolveScope } from "../helpers.ts";
+ 
+@@ -78,7 +79,9 @@ export const checkCmd = new Command()
+           console.error(`error: ${filePath}: file not found`);
+           Deno.exit(1);
+         }
+-        if (filePath.endsWith(".md")) mdContents.set(filePath, content);
++        if (MARKDOWN_EXTENSIONS.has(extname(filePath).toLowerCase())) {
++          mdContents.set(filePath, content);
++        }
+         const result = await parseFile(content, { file: filePath });
+         allEntries.push(...result.entries);
+         parseDiagnostics.push(...result.diagnostics);
+@@ -120,7 +123,6 @@ export const checkCmd = new Command()
+           loadMarkdownFormatter,
+           parseLockfile,
+         } = await import("../../core/mod.ts");
+-        const { join } = await import("@std/path");
+ 
+         const formatMarkdownProse = await loadMarkdownFormatter();
+ 
+diff --git a/packages/markspec/core/formatter/mod.ts b/packages/markspec/core/formatter/mod.ts
+index de801d22..fdfe9121 100644
+--- a/packages/markspec/core/formatter/mod.ts
++++ b/packages/markspec/core/formatter/mod.ts
+@@ -6,6 +6,7 @@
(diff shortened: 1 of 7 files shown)


stasson:
## /review findings (4 finder angles → verified → fixed in `f460a34`)

The three core fixes verified clean — correctness and cross-file tracers found nothing (the try/catch flows, the `polished` undefined-guard, the in-`format()` gate, and the `join` import consolidation all check out; no production caller loses the pass; no import cycle). The cleanup + altitude angles surfaced three diagnostics-quality items on the code this PR touches; all three are addressed in the follow-up commit:

1. **Prose-pass throw reused the gate-rejection's message.** *(both cleanup + altitude flagged)* A dprint throw in `formatProseSegments` routed into the same `fallbackStarts` bucket as a semantic-equivalence rejection, so `format()` emitted "Markdown pass produced non-equivalent output…" even when the formatter had actually thrown — a wrong root-cause lead, and asymmetric with the entry-body path which already distinguished "errored" from "non-equivalent". **Fixed:** threaded a per-fallback `reason` (`"error" | "non-equivalent"`) through `ProsePassResult`; the prose diagnostic now names the accurate cause, matching the body path.

2. **The prose-throw test didn't assert the diagnostic.** It only checked the prose survived, so a regression that dropped or mis-coded the MSL-F012 for prose throws would have shipped green. **Fixed:** the test now asserts an MSL-F012 with the "errored" cause, plus a new `prose_test.ts` case pinning `reason === "error"` for a throwing formatter (and the existing rejection test now pins `reason === "non-equivalent"`).

3. **The Markdown-only gate keys off `options.file`, silently disabling the pass when `file` is absent** — undocumented on an option whose JSDoc said only "for diagnostic messages". Behavior is intended (that's the point of finding #2), but a future JSR consumer passing in-memory markdown + the callback without a `.md` file would get a silent no-op. **Fixed:** documented on `formatMarkdownProse` that the pass is Markdown-only and requires a `.md`-suffixed `file`.

**Deliberately deferred to #669** (loader/diagnostics hardening): capturing the raw `err` text from the two catch blocks. The `reason` tag already tells a triager which failure *class* occurred (throw vs rejection) — enough to pick the right debugging path — without threading untrusted, possibly-multiline WASM error strings through the result type into user-facing diagnostics for a path that has never fired. Richer dprint-failure UX belongs with #669's loader work.

Full `just check` green on both commits.

PR #679, opened 2026-07-02
fix(core): skip fenced code regions in ref canonicalisation (#668)

### LINKED ISSUE #668: fmt: ref-canonicalisation rewrites trace values inside fenced code examples

stasson:
Closes #668.

## Summary

`markspec fmt`'s ADR-026 reference canonicalisation (`canonicalizeRefs`) re-detected trace trailers by regex, line-by-line, with **no fence-awareness** — so an illustrative `Satisfies: 01HGW…`/`Realizes: 01HGW…` line inside a ```` ```markdown ```` / ```` ```text ```` example got "healed" against the real project ref index. Bare `markspec fmt` dirtied 3 spec/ADR docs on every run (`check`'s MSL-F010 gate calls `format()` without a ref index, so CI stayed green — only write-path runs churned).

## Fix

`canonicalizeRefs` now tracks fenced-code regions with the shared `FENCE_RE` and skips trailer rewriting inside them — the same verbatim-region rule every other fmt pass already honors (§5.1; cf. `collapseBlankLines`). ~10 lines, no behavior change outside fences. The multi-line `\`-continuation walk is unaffected: fence markers never match `CONTINUATION_LINE_RE`, so they're always left for the outer loop's toggle.

## Verification

- Two colocated unit tests (both confirmed RED with the guard disabled): a fenced trailer stays byte-for-byte while an identical *real* trailer outside the fence is still healed; a fence-only document is a strict no-op.
- **Reproduce-clean:** bare `markspec fmt` on this repo no longer touches the 3 docs (`docs/architecture/adr-002-entry-model.md`, `docs/spec/internal/markspec-core-data-model.md`, `docs/spec/model/prose-analysis.md`) — the diff is now empty.
- Full `just check` green.

## Scope note

The issue also mentions indented code blocks parenthetically. The repo's own docs put examples in fenced blocks (confirmed: all 3 churned files use ```` ```text ````), so fenced-region handling resolves the reported symptom completely. Indented-code is deliberately left out of scope here: MarkSpec trailers are themselves 4+-space indented, so reliably distinguishing a prose indented-code block from a real trailer needs entry-boundary awareness — a separate, subtler change with no current reproducer.

### DIFF
--- a/.pr679.diff
+++ b/.pr679.diff
@@ -1,0 +1,118 @@
+diff --git a/packages/markspec/core/refs/mod.ts b/packages/markspec/core/refs/mod.ts
+index 13e01ffc..66586119 100644
+--- a/packages/markspec/core/refs/mod.ts
++++ b/packages/markspec/core/refs/mod.ts
+@@ -20,6 +20,7 @@ import {
+   ULID_RE,
+ } from "../model/mod.ts";
+ import type { LockEdge } from "../lock/mod.ts";
++import { FENCE_RE } from "../util/fence.ts";
+ 
+ /**
+  * Trace-attribute keys whose values are canonicalised/healed: the core
+@@ -112,12 +113,24 @@ export function canonicalizeRefs(
+ 
+   const lines = content.split("\n");
+   let changed = false;
++  // Track fenced-code regions so illustrative trailers inside ``` / ~~~
++  // blocks (e.g. `Realizes: 01HGW...` in a ```markdown example) are never
++  // "healed" against the real ref index — the same verbatim-region rule
++  // every other fmt pass honors (§5.1; cf. `collapseBlankLines`). Without
++  // this, `markspec fmt` rewrites example IDs in spec/ADR docs on every run
++  // (#668). The parser/validator already treat fenced content as inert.
++  let inFence = false;
+   for (let i = 0; i < lines.length; i++) {
+     // Strip a trailing CR before matching so CRLF files are handled (the
+     // trailer regex ends in `$`, which `\r` would block) and re-append it on
+     // rewrite so line endings stay byte-for-byte lossless.
+     const cr = lines[i].endsWith("\r");
+     const bare = cr ? lines[i].slice(0, -1) : lines[i];
++    if (FENCE_RE.test(bare)) {
++      inFence = !inFence;
++      continue;
++    }
++    if (inFence) continue;
+     const m = TRACE_TRAILER_RE.exec(bare);
+     if (!m) continue;
+     const [, indent, key, sep, rest] = m;
+diff --git a/packages/markspec/core/refs/mod_test.ts b/packages/markspec/core/refs/mod_test.ts
+index 3016fb9c..308ef6b4 100644
+--- a/packages/markspec/core/refs/mod_test.ts
++++ b/packages/markspec/core/refs/mod_test.ts
+@@ -145,6 +145,76 @@ Deno.test("canonicalizeRefs: ULID token in trace attr replaced with current disp
+   );
+ });
+ 
++// ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


stasson:
## /review findings (fence-guard correctness + cross-file/reuse/conventions → verified)

The core fix is sound and empirically proven (bare `markspec fmt` on this repo goes from 3 churned docs to zero). Review surfaced one genuine robustness hole in the fix itself (now fixed in `059896e`), one out-of-scope sibling bug (filed as a follow-up), and three shared-`FENCE_RE` limitations that are consistent with the rest of the codebase and cause no active regression.

### Fixed in this PR (`059896e`)

- **`core/refs/mod.ts` — the `\`-continuation walk could desync fence tracking.** `CONTINUATION_LINE_RE` (`/^(\s+)(\S.*)$/`) matches an *indented* fence marker, so a `\`-continued trailer flowing straight into an indented ```` ```md ```` block would swallow the marker as continuation text without toggling `inFence`, leaving every trailer in that "fence" exposed to healing. Contrived layout (needs no blank line between a continued trailer and an indented fence), but a real desync once hit. Fixed: the continuation walk now backs out on a `FENCE_RE` line and lets the outer loop toggle it, with a RED-verified regression test.

### Filed as follow-up — [#680](https://github.com/driftsys/markspec/issues/680) (out of scope for #668)

- **LSP workspace rename has the identical fenced-example blind spot.** `findIdOccurrencesInFile` (`lsp/rename.ts`) whole-token-replaces a display ID across raw text with no fence awareness, so renaming a real entry rewrites the ID inside illustrative fenced examples too. Same class as #668 but a different subsystem (and a previewable, user-initiated action), so it's a separate fix, not a blocker here.

### Accepted limitations (shared with every fence-aware pass; no active regression)

These are properties of the shared `FENCE_RE` flat toggle that `walkProseLines`/`collapseBlankLines` already rely on — making `canonicalizeRefs` stricter than the rest would be inconsistent, and the whole-repo reproduce-clean confirms none of them churns a real doc today:

- **Nested fences** (a 4-backtick block wrapping a 3-backtick block, an idiom this repo's typl docs use) mis-toggle `inFence` — latent, not active (reproduce-clean is clean).
- **Mismatched marker** (a ```` ``` ````-opened block containing a `~~~` line) toggles early — same flat-toggle root.
- **Unterminated fence** leaves `inFence` true to EOF, skipping subsequent trailers — but per CommonMark an unclosed fence *is* code to EOF, so not healing those is correct, not a regression. (A proper fix for all three would be a CommonMark-correct, length/char-aware `FENCE_RE` shared across every pass — a separate change, deliberately not gold-plated here.)

Full `just check` green on both commits.

PR #681, opened 2026-07-02
fix(core): harden dprint loader — freeze configs, clean load error, LSP fallback visibility (#669)

### LINKED ISSUE #669: fmt/lsp: dprint-markdown loader hardening (frozen configs, clean load-failure error, LSP MSL-F011 visibility)

stasson:
Closes #669.

Three small hardening items deferred from the ADR-029 (#649) reviews — none was a correctness bug; all improve robustness/UX around the embedded dprint-markdown loader.

## Changes

1. **Freeze the exported style constants** (`core/formatter/dprint.ts`). `MARKSPEC_MARKDOWN_GLOBAL_CONFIG` / `MARKSPEC_MARKDOWN_PLUGIN_CONFIG` are exported through the public barrel and baked into the formatter once at `instantiate()`. They're now `Object.freeze`d (and typed `Readonly<…>`) so a consumer can't silently mutate the canonical style out from under the running formatter (which would drift the entry-body width budget). Unit test asserts both are frozen.

2. **Clean error on WASM load failure** (`cli/commands/fmt.ts`, `cli/commands/check.ts`). `await loadMarkdownFormatter()` was unguarded — a corrupt/missing plugin surfaced as a raw stack trace. Both now catch and emit the CLI's standard `error: could not load the Markdown formatter: <msg>` + `Deno.exit(1)`. Fails before any file is written (the loader is awaited once, before the write loop).

3. **LSP fallback visibility** (`lsp/server.ts`). `onDocumentFormatting` discarded `result.diagnostics`, so a Markdown-pass fallback (**MSL-F012**, info — note: this code was F011 in the issue title, renamed to F012 when v0.9.0's ref-canon-drift gate took F011) was silent in the editor while the CLI prints it. The handler now emits one `window/logMessage` summary line via `connection.console.info` when any F012 fallback occurred — visible in the editor's output channel, no toast-per-save.

## Testing note

Item 1 has a deterministic unit test. Items 2 and 3 are defensive/observability changes on paths that require a genuine WASM-load failure (item 2) or a real dprint semantic-gate fallback (item 3) to trigger — neither is deterministically reproducible without adding an injection seam that would be more machinery than the fix. They follow the file's existing untested patterns for `console.error`+exit and `connection.console.*` logging. Full `just check` green.

### DIFF
--- a/packages/markspec/cli/commands/check.ts
+++ b/packages/markspec/cli/commands/check.ts
@@ -124,7 +124,21 @@ export const checkCmd = new Command()
           parseLockfile,
         } = await import("../../core/mod.ts");
 
-        const formatMarkdownProse = await loadMarkdownFormatter();
+        let formatMarkdownProse: Awaited<
+          ReturnType<typeof loadMarkdownFormatter>
+        >;
+        try {
+          formatMarkdownProse = await loadMarkdownFormatter();
+        } catch (err) {
+          // The embedded dprint-markdown plugin could not be loaded (corrupt
+          // or missing WASM, restricted filesystem). Fail with a clean
+          // message rather than a raw stack trace.
+          const msg = err instanceof Error ? err.message : String(err);
+          console.error(
+            `error: could not load the Markdown formatter: ${msg}`,
+          );
+          Deno.exit(1);
+        }
 
         // Read markspec.lock once: its edge ledger feeds reference healing
         // (MSL-F011) and its cached edge hash feeds the lockfile gate
--- a/packages/markspec/cli/commands/check.ts
+++ b/packages/markspec/cli/commands/check.ts
@@ -8,7 +8,12 @@ import { extname, join } from "@std/path";
 import { Command } from "@cliffy/command";
 import { ConfigError, MARKDOWN_EXTENSIONS } from "../../core/mod.ts";
 import type { CaptionConventions, Diagnostic } from "../../core/mod.ts";
-import { loadActiveProfile, readFile, resolveScope } from "../helpers.ts";
+import {
+  loadActiveProfile,
+  loadMarkdownFormatterOrExit,
+  readFile,
+  resolveScope,
+} from "../helpers.ts";
 
 export const checkCmd = new Command()
   .description("Check broken refs, missing Ids, duplicates")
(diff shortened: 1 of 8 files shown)


stasson:
## /review findings (3 finder angles → verified → applied in `dc35513`)

The three original changes verified correct — the cross-file finder confirmed via the `@dprint/formatter` source that `setConfig` only JSON-stringifies its args (so the freeze is safe), that the only two config read-sites are read-only, and that `check.ts` scope + the `const→let` + `Deno.exit(): never` control flow are all sound. No correctness bugs. Three improvement findings surfaced, all applied:

1. **Duplicated load-or-exit try/catch** (fmt.ts + check.ts) — 2-use duplication that already fits the `cli/helpers.ts` convention (`requireProjectConfig`, `loadActiveProfile`). **Applied:** extracted `loadMarkdownFormatterOrExit()` into `cli/helpers.ts`; both commands now call it (−~20 lines, single source for the error message).

2. **Inline, untested LSP F012 count+message** — the same file already extracts analogous logic (`buildDiagnosticsHistogram`) into a pure, tested helper. **Applied:** extracted `buildF012FallbackMessage(diagnostics): string | undefined` into `lsp/formatting.ts` with unit tests (counts only MSL-F012, names the count, `undefined` when none) — this also closes the item-3 test gap noted in the PR body.

3. **LSP loader await was unguarded** (correctness finder). The cross-file finder verified vscode-jsonrpc already wraps a handler rejection in a clean `ResponseError` (not a raw stack trace), and #669 item 2 scoped the clean-error to the CLI — so this wasn't the raw-stack-trace failure the finder first described. But graceful degradation is strictly better than an error response: **applied** — on WASM-load failure the handler now logs and falls back to **entry-only formatting** (ULID stamping, trailer normalization still apply), so format-on-save keeps working; the loader un-caches the failure so a later request retries.

Full `just check` green on both commits (77 formatter unit + 61 fmt/check/lsp e2e + the new formatting_test cases).

Issue #657, opened 2026-07-01
discovery: bare check/lint/fmt no longer skip node_modules/target/dist/build by default


stasson:
## Context

Review follow-up from PR #651 (finding #4) — see [review comment](https://github.com/driftsys/markspec/pull/651#issuecomment-4859262420). Left out of scope of the merged PR because it needs a design call.

## Problem

The unified `core/discovery` walker that PR #651 introduced skips, in precedence order: (1) hidden directories (name starts with `.`), (2) `.gitignore` rules, (3) `project.yaml` `exclude:` patterns.

- `packages/markspec/core/discovery/mod.ts` — `walk()` only hard-skips `entry.name.startsWith(".")`; everything else depends on gitignore/exclude.

The **previous** LSP walker had a built-in skip set:

```
node_modules, .git, .worktrees, target, dist, build, skills
```

So on a project that does **not** gitignore these (or lacks a `.gitignore` entirely), bare `check`/`lint`/`fmt` will now descend into `node_modules/`, `target/`, `dist/`, `build/` — scanning vendored/generated Markdown that was previously skipped for free. (`.git`/`.worktrees` are still skipped as hidden dirs; `skills/` is handled per-project via `exclude:`.)

## Decision needed

1. **Reintroduce a built-in skip set** in `core/discovery` for well-known build/vendor dirs (`node_modules`, `target`, `dist`, `build`), applied before gitignore. Cost: a project that *wants* to scan such a dir must opt back in.
2. **Document the `.gitignore` dependency** — declare that discovery honors `.gitignore` and require projects to gitignore their build/vendor dirs (the common case already does).

## Acceptance

- The chosen behaviour is implemented and covered by a `core/discovery` unit test (a project with a non-ignored `node_modules/x.md`).
- Behaviour documented in ADR-027 / the CLI reference.

stasson:
Already resolved on `main` — this is a duplicate of #661, which shipped via `8b578de` (in v0.10.0).

The chosen resolution matches Option 1 in this issue: `core/discovery` restores a built-in skip for `node_modules`, `target`, `dist`, `build` (`BUILTIN_SKIP_DIRS`), applied as the **lowest-precedence** rule layer — before `.gitignore` and `exclude:` — so a `!target/` negation in either re-includes them under gitignore last-match-wins. One SSOT; every discovery consumer (CLI verbs, LSP, lockfile) inherits it.

**Re-verified on current main today:**
```
[...]

Issue #656, opened 2026-07-01
check: MSL-F010 drift gate does not run canonicalizeRefs — check/fmt disagree on "formatted"


stasson:
## Context

Review follow-up from PR #651 (finding #2) — see [review comment](https://github.com/driftsys/markspec/pull/651#issuecomment-4859262420). Deliberately left out of scope of the merged PR because it needs a design decision.

## Problem

`markspec check`'s fmt-drift gate (MSL-F010) runs only the structural formatter:

- `packages/markspec/cli/commands/check.ts` — the MSL-F010 block calls `format(content, { file })` and flags the file when `.changed` is true.

But `markspec fmt` does **two** things over a project: `format()` **and** `canonicalizeRefs()` (ADR-026 trace-ref canonicalisation/healing, `core/refs`). Because `check` runs only `format()`, a file that is structurally canonical but still carries an un-canonicalised trace reference (e.g. a ULID that `fmt` would rewrite to a display ID, or a stale display ID `fmt` would heal via the lockfile ledger) is **not** reported as drift by MSL-F010. `check` and `fmt` therefore disagree about what "formatted" means.

## Decision needed

Pick one and make it explicit:

1. **Run canon in `check`** — have the MSL-F010 gate also apply `canonicalizeRefs()` (with the same project ref-index + lockfile ledger `fmt` uses) so `check` catches ref drift too. Cost: `check` must build the ref index (project-wide) like `fmt` now does.
2. **Document format-only scope** — declare MSL-F010 as structural-format drift only, and rely on `markspec fmt` (or a separate ref-drift diagnostic) for canonicalisation. Cost: a canonical-looking file can still need `fmt`.

## Acceptance

- The chosen behaviour is implemented and covered by an e2e test (a project whose only drift is an un-canonicalised trace ref).
- The scope of MSL-F010 is documented (guide/CLI reference + ADR-027 note).

stasson:
Already resolved on `main` — this is a duplicate of #660, which shipped via PR #662 (`aa3f89a`, in v0.10.0).

The chosen resolution matches Option 1 in this issue: `check`'s fmt-drift gate now runs the same `format() → parse → canonicalizeRefs` sequence `fmt` does, emitting a distinct **MSL-F011** for reference-canonicalization drift (a ULID or stale display ID `fmt` would rewrite), kept separate from MSL-F010. Project-wide only; `check <file>` never fires it.

**Re-verified on current main today:**
```
[...]

PR #682, opened 2026-07-02
feat(cli): surface lockfile edge-drift in doctor (#658)

### LINKED ISSUE #658: check: soften MSL-L212 lockfile-drift hard error on first-run/upgrade
## Context

Review follow-up from PR #651 (finding #5) — see [review comment](https://github.com/driftsys/markspec/pull/651#issuecomment-4859262420). Left out of scope of the merged PR.

## Problem

`markspec check`'s offline lockfile-drift gate emits **MSL-L212** at `severity: "error"`, which blocks `check` (exit 1):

- `packages/markspec/cli/commands/check.ts` — the lockfile block pushes MSL-L212 with `severity: "error"` when the current canonical edge hash differs from `markspec.lock`.

This is correct steady-state behaviour, but it makes the **first upgrade** hostile: a project that adopts this MarkSpec version and already has a `markspec.lock` (or generates one, then edits the graph) gets a hard blocking error until it runs `markspec lock`. For teams pulling the upgrade, `check` goes red in CI with no gentle onramp.

## Options

1. **Warn-on-first-run** — emit MSL-L212 as a `warning` (exit 2, non-blocking) the first time drift is seen after an upgrade, promoting to `error` once acknowledged / on subsequent runs. (Needs a marker for "acknowledged".)
2. **Release note only** — keep it a hard error but call out the `markspec lock` step prominently in the upgrade/release notes and `doctor` output.
3. **`--strict`-gated** — demote MSL-L212 to a warning by default, promote to error under `--strict` (consistent with how prose findings already work in `check`).

Note: project policy forbids *persistent* demotion of real errors ([[project_strict_relax_mode]] context) — favour option 2 or a bounded first-run grace, not a permanent downgrade.

## Acceptance

- Chosen UX implemented + e2e covered.
- Upgrade path documented (release notes / `doctor`).

### LINKED ISSUE #660: check: MSL-F010 gate misses ref-canonicalization drift (false all-clean)
Follow-up from the PR #651 review (finding #2) and ADR-027 Follow-ups.

## Problem

`markspec check`'s project-wide fmt-drift gate (`cli/commands/check.ts`) runs only `format(content).changed`. It does **not** run the `canonicalizeRefs` pass that `markspec fmt` also performs (`cli/commands/fmt.ts`). So bare `check` exits 0 on a file that `fmt` / `fmt --check` would still rewrite — a non-canonical but formatter-clean display-ID/ULID reference. The composite gate's "fmt-check" is therefore incomplete: CI can report a false all-clean.

False-negative only (no false positives).

## Decision (already made in review — Option A)

Make `check` detect canonicalization drift too, using `fmt`'s exact core functions so drift detection can't diverge from what `fmt` writes:

- In `check`'s project-wide gate region, build a ref index from the already-available exclude-aware `allEntries` (`buildRefIndex`) plus the lockfile edge ledger already read by the L212 gate, then for each markdown file run the same `format(content) -> parse -> canonicalizeRefs(...)` sequence and detect drift.
- Emit a **new distinct code `MSL-F011`** ("references are not canonical - run `markspec fmt`"), error severity, **project-wide only** (guard with `scope.projectWide`, same as `MSL-F010`). Do not fold into `MSL-F010` so the author knows which fmt concern fired.

Accepted consequence: bare `check` now hard-fails on canonicalization drift too. It is the CI gate; the canonical `insert -> fmt -> check` agent loop runs `fmt` first. Documented in ADR-027 + release notes.

## Docs (same PR)

Document `MSL-F010`, the new `MSL-F011`, and the in-`check` `MSL-L212` in the governed diagnostic catalogue (`docs/spec/language/language.md` §8), per ADR-012's phased-adoption process (coordinate with ADR-012; do not invent a scheme).

## Acceptance

- A project where a real reference is non-canonical but formatter-clean: bare `check` exits 1 with `MSL-F011`; after `markspec fmt`, `check` exits 0.
- File-local `check <file>` does not fire `MSL-F011` (project-wide only).
- `MSL-F010`, `MSL-F011`, `MSL-L212` documented in language.md §8.

stasson:
Closes #658.

## What

`markspec check`'s `MSL-L212` gate hard-fails (exit 1) on lockfile edge-drift. That makes the **first `check` after a MarkSpec upgrade** hostile — a red CI build with no heads-up, cleared by one `markspec lock`. Project policy forbids persistently demoting a real error, so this surfaces the *same* drift proactively in `markspec doctor` as a **non-blocking warning**, while `check` keeps the hard error.

## How

- **One shared drift primitive (SSOT).** Factored the offline edge-hash comparison out of `check.ts`'s inline `MSL-L212` block into a pure helper, `detectOfflineEdgeDrift(projectEntries, generatedCache)` in `core/lock/edge_drift.ts`. Both `check` (refactored — message + severity + location unchanged) and `doctor` (new) call it, so they can never disagree on what "drift" means — the failure mode #656 was about.
- **`doctor` lock-drift health check.** Runs only when a parseable `markspec.lock` exists (skips the project walk otherwise, so `doctor` stays fast). Discovers + parses the project (corpus-blind `!e.origin`, matching `check`), and on drift emits `lockfile-edge-drift` — a non-`MSL` `doctor` code (like `toolchain-below-floor`, deliberately outside the governed catalogue / ADR-012 migration), a warning contributing to **exit 2**. Adds a text status line and a `lockfile` block to `--format json`.

```text
⚠ Lockfile: traceability edges drifted (locked 12, current 15) — run `markspec lock`
```

## Tests

- **Unit** — `detectOfflineEdgeDrift` (in-sync vs. changed graph).
- **E2E** — `doctor_lock_drift_test.ts`: in-sync → clean exit 0 · drift → warn exit 2 + `lockfile-edge-drift` + JSON `edgeDrift:true` · no lockfile → `present:false`, exit 0. Uses a real lockfile via `markspec lock`, then mutates the graph.
- **Fixture** — the toolchain-skew tests' shared `lockfile()` helper now uses a real empty-edge hash; a placeholder would (correctly) trip the new drift check and mask the floor behaviour they target.
- **Regression** — existing `check: lockfile edge drift fails with MSL-L212` stays green after the refactor. Full suite: **3453 passed, 0 failed**.

## Docs

`docs/guide/cli.md` (doctor section + `lock` upgrade note) and ADR-027 (the shared-helper parity note). No `CHANGELOG` entry — batched at release.

## Also in this session

#656 and #657 were found to be already implemented, tested, and documented on `main` (duplicates of the closed #660/#661), and were re-verified end-to-end and closed. This PR is the only remaining work of the three.


stasson:
## `/review` — 4 independent finder angles (correctness · removed-behavior/cross-file · cleanup/altitude · conventions), 1-vote verify

**Overview.** The PR extracts `check`'s inline `MSL-L212` edge-hash comparison into a pure `detectOfflineEdgeDrift` helper that both `check` (behaviour-preserving refactor) and the new `doctor` drift check call, and surfaces lockfile drift in `doctor` as a non-blocking warning. Verified byte-identical `MSL-L212` message/severity/location, no export collisions, no dangling imports, and that `doctor`'s discovered file set matches `markspec lock`'s pin. No correctness bugs or convention violations found.

Three findings surfaced (two of them independently by 3 of the 4 finders). Disposition:

### Fixed in this PR

1. **`doctor.ts` — malformed-lockfile JSON asymmetry** (correctness + cleanup angles). A present-but-malformed `markspec.lock` reported `--format json` `lockfile: { present: true, edgeDrift: false, lockedEdges: 0, currentEdges: 0 }` while the text output (guarded by `if (parsedLock)`) correctly printed nothing — a JSON/CI consumer would read `edgeDrift: false` as "in sync" when no comparison ran. **Fix:** the JSON `lockfile` block now emits the drift fields only when the lockfile actually parsed (`parsedLock`), mirroring the text guard; a malformed lockfile reports `{ present: true }` with the drift fields omitted. Covered by a new e2e (`present-but-malformed lockfile → present:true, no drift fields, exits 0`) and documented in cli.md.

2. **`doctor.ts` — redundant `extensions: RELEVANT_EXTENSIONS` arg** (cleanup angle). `discoverFiles` already defaults `extensions` to `RELEVANT_EXTENSIONS`, and `lock`'s `collectEntries` relies on that default — pinning it explicitly in `doctor` would split the two edge collections (breaking drift parity) if the default ever changed. **Fix:** dropped the explicit arg + the now-unused import, so `doctor` relies on the same default `lock` does.

### Deferred → #684

3. **Three duplicated project-entry collectors** (cleanup/altitude angle). `lock`'s `collectEntries` (the pinner), `check`'s inline loop, and `doctor`'s new loop must all stay byte-identical for the drift comparison to be meaningful. Real maintainability risk, but a cross-cutting refactor beyond this S-issue — filed as #684 (extract one shared, tested collector consumed by all three).

Conventions (AGENTS.md/CLAUDE.md): clean — no `Deno.*` in the new core file, `core/mod.ts` boundary respected, no cycles, public API documented, `lockfile-edge-drift` matches the existing non-`MSL` `toolchain-below-floor` code form.

Full suite after fixes: **3454 passed, 0 failed**.

PR #683, opened 2026-07-02
fix(core): compute prose-lint scope with profile and corpus-awareness

### LINKED ISSUE #675: Prose lint never runs on profile-typed entries (isProseScope is profile-blind)

stasson:
Closes #675. Addresses #674 findings 1, 2, and 4 (finding 3 — LSP raw cache-path leak — is not touched here).

## Problem

One coherent defect family: **prose-lint scope was computed at the wrong layer / with the wrong inputs**, producing one silent miss and two silent false positives.

- **#675** — `isProseScope` called `resolvedCoreType(entry)` **without** the active profile, so entries classified purely by a profile type (via `Type:` or a profile `display-id-pattern`, e.g. `platform-component`) never resolved to a core `Specification` descendant and never entered prose scope. `markspec lint` and `check`'s advisory gate silently skipped most profile-typed requirements.
- **#674 finding 1** — `check` pre-filtered corpus entries **out of `runLint`'s input**, breaking the MSL-Q500 additive-index contract: a `$Identifier` defined only in a delivered-corpus doc no longer silenced Q500, so a project entry referencing it tripped a spurious `xref-glossary-undefined` warning.
- **#674 finding 2** — standalone `markspec lint` passed the merged entry set (now including corpus) into `runLint` with **no origin filter**, so prose analysis ran on delivered-corpus entries (ADR-030 §D4 violation): skewed `score` roll-ups and could exit 1 on unfixable upstream prose under `--strict`.
- **#674 finding 4** — `compileProject` printed `corpus.diagnostics` to stderr but dropped them from the returned `CompileResult.diagnostics`, so `export`/`compile --format json` omitted warning-severity corpus-load issues (e.g. `PROFILE-DELIVERS-002`) that `check` and MCP both report.

## Fix

The corpus policy now lives in one place — `runLint`:

- Thread `EffectiveProfile` through `LintOptions` → `runLint` → `isProseScope` → `resolvedCoreType`, so the profile's `extends:` chain resolves profile-typed entries into scope (#675).
- Build the glossary / `$Identifier` resolver indexes from the **full** entry set (corpus included, additive), but exclude delivered-corpus entries (`entry.origin`) from rule execution, suppression hygiene, and the score roll-up. Both call sites now pass the merged project + corpus set plus the profile — no more input pre-filter (#674 f1/f2).
- `compileProject` merges `corpus.diagnostics` into the returned `CompileResult.diagnostics` so serialized artifacts surface them too (#674 f4). Only warning/info corpus diagnostics reach the merge — error-severity ones already exit first; both sets were already printed to stderr, so this changes serialization only.

## Tests

- 4 e2e (`tests/e2e/delivered_test.ts`): profile-typed entry fires MSL-Q300; a corpus `$Identifier` silences Q500 in project prose; standalone `lint` never lints corpus but still lints project prose; `export json` carries the corpus-load warning. Each was confirmed RED before the fix.
- 2 runner unit tests (`runner_test.ts`): `isProseScope` needs the profile to scope a profile-typed entry; a corpus entry silences Q500 yet emits no diagnostics and is absent from the score roll-up.
- `just build` green (lint + test + typecheck + compile); `deno fmt --check` + `dprint check` clean.

## Review disposition

Two independent review passes surfaced only non-blocking items:

- **Declined (out of scope):** merged corpus-load diagnostics carry an absolute `location.file`. This is consistent with how `check --format json` already emits corpus diagnostic locations and does not break the tested same-machine determinism contract; unifying serialized corpus-diagnostic location representation across `check` and `export` is a separate decision. Worth a follow-up.
- **Declined (out of scope):** the `entry.origin` corpus test is open-coded across ~6 existing sites; no `isCorpusEntry` helper exists. Centralizing it is an unrelated cross-cutting refactor.

### DIFF
--- a/packages/markspec/cli/commands/check.ts
+++ b/packages/markspec/cli/commands/check.ts
@@ -254,10 +254,14 @@ export const checkCmd = new Command()
       const proseDiagnostics: Diagnostic[] = [];
       if (scope.projectWide) {
         const { runLint } = await import("../../core/mod.ts");
-        // Prose lint never runs on delivered corpus entries (ADR-030) — a
-        // consumer cannot fix an upstream profile's prose.
+        // Pass the FULL entry set (project + corpus) and the active profile.
+        // `runLint` excludes corpus entries from its output but keeps them in
+        // the glossary / $Identifier indexes, so a corpus-defined term still
+        // silences Q500 in project prose (ADR-030 §D4). The profile threads
+        // into `isProseScope` so profile-typed entries are scoped (#675).
         const lintResult = await runLint({
-          entries: allEntries.filter((e) => !e.origin),
+          entries: allEntries,
+          profile: chain?.effective,
         });
         for (const d of lintResult.diagnostics) {
           proseDiagnostics.push({
--- a/packages/markspec/cli/commands/lint.ts
+++ b/packages/markspec/cli/commands/lint.ts
@@ -28,10 +28,15 @@ export const lintCmd = new Command()
         verb: "linting",
         quiet: options.quiet === true || options.format === "json",
       });
-      const { result } = await compileProject(scope.files);
+      const { result, chain } = await compileProject(scope.files);
       const { runLint } = await import("../../core/mod.ts");
+      // `result.entries` is the merged project + delivered-corpus set. Pass it
+      // whole with the active profile: `runLint` keeps corpus entries in its
+      // resolver indexes but never emits diagnostics for them (ADR-030 §D4),
+      // and the profile scopes profile-typed entries into prose analysis (#675).
       const lintResult = await runLint({
         entries: [...result.entries.values()],
+        profile: chain?.effective,
       });
 
       let diagnostics = [...lintResult.diagnostics];
--- a/packages/markspec/cli/helpers.ts
+++ b/packages/markspec/cli/helpers.ts
@@ -255,7 +255,20 @@ export async function compileProject(
     );
   }
 
-  return { result, chain };
+  // Merge the corpus-load diagnostics into the returned result so
+  // serialized artifacts (`export`/`compile --format json`) surface them
+  // too — otherwise a machine consumer parsing the JSON would believe the
+  // corpus is healthy while other surfaces (`check`, MCP) flag it (#674 f4).
+  // Only warning/info corpus diagnostics reach here: error-severity ones
+  // already exited above. Prepended so they read as load-time findings,
+  // ahead of the compiler's own diagnostics. Both sets were already printed
+  // to stderr above, so this changes serialization only, never the console.
+  const merged: CompileResult = {
+    ...result,
+    diagnostics: [...corpus.diagnostics, ...result.diagnostics],
+  };
+
+  return { result: merged, chain };
 }
 
 /**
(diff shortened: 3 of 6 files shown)


stasson:
## Review (high-effort, two independent passes)

A correctness pass (line-by-line + removed-behavior + cross-file) and a quality pass (reuse / simplification / efficiency / altitude / conventions) were run over the diff. Summary below; nothing blocking survived verification.

**Verified clean:**
- `entry.origin` is `EntryOrigin | undefined` (always an object or absent) — `!entry.origin` is an exact project-vs-corpus test with no falsy-but-defined trap. `loadDeliveredCorpus` stamps `origin` on every corpus entry.
- The three corpus skips in `runLint` (rule loop, hygiene loop, score roll-up) are consistent; the glossary / `$Identifier` indexes still see the **full** set, so Q500 fires ≤ before, never more (additive-enrichment invariant preserved).
- `computeScoreRollup(out, entries.filter(!origin))`: `out` contains only project-entry diagnostics (corpus skipped in both producing loops) and corpus files are distinct paths — no orphaned or cross-contaminated diagnostics.
- `compileProject` merge: error-severity corpus diagnostics `Deno.exit(1)` **before** the merge (only warning/info reach it); `chain === null` makes the merge a no-op; corpus **load** diagnostics are disjoint from compile-time diagnostics (compiler receives already-parsed `corpusEntries`, never re-loads) — no double-count.
- No exit-code change: all 13 `compileProject` callers were traced; none derive an exit code or stderr print from `result.diagnostics` (`doc.ts` prints the `renderPdf` result, not compileProject's). Only `export`/`compile --format json` serialize it — the intended fix.
- `score_text.ts` (`runLint({ entries: [entry] })`) unaffected by the new optional `profile`.

**Addressed:** redundant `?? undefined` after `chain?.effective` (non-nullable `effective`) removed at both call sites.

**Non-blocking, deliberately out of scope:**
- Merged corpus-load diagnostics carry an absolute `location.file`. This matches how `check --format json` already emits corpus diagnostic locations and does not break the tested same-machine determinism contract. Unifying serialized corpus-diagnostic location representation across `check` and `export` is a separate decision — worth a follow-up.
- The `entry.origin` corpus test is open-coded across ~6 existing sites; no `isCorpusEntry` helper exists. Centralizing it is an unrelated cross-cutting refactor.

PR #690, opened 2026-07-03
refactor(core): single shared project-entry collector (#684)

### LINKED ISSUE #684: refactor(core): single shared project-entry collector for lock/check/doctor edge-hash parity
## Context

Review follow-up from PR #682 (the `doctor` lockfile edge-drift feature, #658). Left out of scope of that PR because it is a cross-cutting refactor.

## Problem

Three independent hand-rolled "discover project files → parse → collect `Entry[]`" loops now exist, and the correctness of `doctor`'s / `check`'s lockfile-drift comparison depends on all three producing the **byte-identical** file set that `markspec lock` pinned:

- `cli/commands/lock.ts` — `collectEntries(root, exclude)` (the *pinner*: computes the edge hash written to `markspec.lock`'s `generated-cache`).
- `cli/commands/check.ts` — inline discovery + parse feeding the `MSL-L212` gate (the *checker*).
- `cli/commands/doctor.ts` — the new inline loop feeding `detectOfflineEdgeDrift` (a second *checker*, added in #682).

They agree today (all go through `core/discovery`'s `discoverFiles` with `config.exclude` and the default `RELEVANT_EXTENSIONS`). But the invariant is enforced only by the three loops staying identical by hand — a future change to one (an added filter, a different `parseFile` option, a divergent exclude source) silently makes a checker report phantom drift or miss real drift relative to what `lock` pinned. #682 already had one near-miss here: `doctor` initially pinned `extensions: RELEVANT_EXTENSIONS` explicitly while `lock` relied on the default, which would have split the two collections if the default ever changed (fixed in-PR by dropping the explicit arg).

## Suggested fix

Extract one shared, exported collector (in `core/`, taking the injected discovery IO) that returns the project entry set, and have the pinner (`lock`) and both checkers (`check`, `doctor`) consume it — so "the set of files whose edges are hashed" has a single definition. When doing so, decide the concurrency policy once (the LSP's `onInitialized` uses bounded concurrency; the CLI loops are serial).

## Acceptance

- One shared collector consumed by `lock`, `check`, and `doctor`; the three inline loops retired.
- Edge-hash parity between `lock` (pinner) and `check`/`doctor` (checkers) covered by a test that would fail if the collectors diverged.

stasson:
Closes #684.

## What

`markspec lock` (the edge-hash **pinner**) and its **checkers** computed "the set of files whose trace edges the lockfile pins" through **three** independent discover→parse loops:

- a private `collectEntries` in `cli/commands/lock.ts` that `compile` and `fmt` reached into (cross-command imports),
- a hand-rolled copy in `doctor` (added in #682).

They agreed only by staying byte-identical **by hand** — a future divergence (an added filter, a different `parseFile` option, a pinned extension set) would make a checker report *phantom* drift, or *miss* real drift, relative to what `lock` pinned. This is the review finding from #682 that #684 tracked.

## How

Extract one exported core helper — `collectProjectEntries(root, io, { exclude })` in **`core/collect/`** — and route `lock`, `compile`, `fmt`, and `doctor` through it. Injected-IO (Node-safe, no `Deno.*`); relies on `discoverFiles`' default extension set (`RELEVANT_EXTENSIONS`) so no caller pins the extensions independently (a pinned value would re-open the divergence — the exact footgun #682's review caught in `doctor`).

**Behaviour-preserving:** same discovery + parse, returns `Entry[]`. `lock`'s private `collectEntries` is deleted; `compile`/`fmt` now import from core instead of reaching into a command module.

**`check` stays on its `resolveScope` loop by design** — it needs each file's raw *content* + per-file *directives* (the fmt-drift `mdContents` and listing contexts), not just entries. Its bare-mode file set is equivalent by construction (same discovery, same `exclude:`, same default extensions) — documented inline and in ADR-027.

## Tests

- **Unit** — `core/collect/mod_test.ts`: `collectProjectEntries` discovers + parses across a tree, honors `exclude`, and relies on the default extension set (ignores `.txt`).
- **Regression** — the `lock` / `compile` / `fmt` / `doctor` e2e are the behaviour-preserving guard (70 affected pass, incl. the lock↔doctor and lock↔check in-sync/drift parity). Full suite: **3488 passed, 0 failed**.

## Docs

ADR-027 "Lock ↔ check edge-hash parity" section documents the single collector and the `check` exception; net **−53 lines** from removing the duplicated loops.

## Rider fix (surfaced by review)

`compile --frozen` now passes `project.yaml` `exclude:` to the collector. It previously collected with no `exclude`, so a project excluding an entry-bearing path (e.g. this repo's `skills/`) would spuriously drift the frozen check (`MSL-L212`) against a lockfile that never counted those entries. New e2e `compile_frozen_exclude_test.ts` guards it (RED before, GREEN after). Full suite: **3507 passed**.

### DIFF
--- a/docs/architecture/adr-027-cli-smoother-defaults.md
+++ b/docs/architecture/adr-027-cli-smoother-defaults.md
@@ -153,16 +153,29 @@ cross-file target).
 
 ### Lock ↔ check edge-hash parity
 
-`markspec lock`'s entry collection (`cli/commands/lock.ts` `collectEntries`) and
-`check`'s `MSL-L212` gate both walk the file set via `core/discovery` with the
-same `RELEVANT_EXTENSIONS` + the same `exclude:` patterns. This is load-bearing:
-the lockfile gate recomputes the canonical edge hash from the
-currently-discovered corpus and compares it to the locked hash — if `lock` and
-`check` walked different file sets, the hash would drift spuriously on every
-run. Unifying both on `core/discovery` (rather than `lock.ts` keeping its own
-markdown-only `walkMarkdown`) is what makes the parity hold; it also widened
-`lock`'s collection to include source-file entries (a lockfile generated before
-this change needs one `markspec lock` refresh).
+`markspec lock`'s entry collection and `check`'s `MSL-L212` gate both walk the
+file set via `core/discovery` with the same default `RELEVANT_EXTENSIONS` + the
+same `exclude:` patterns. This is load-bearing: the lockfile gate recomputes the
+canonical edge hash from the currently-discovered corpus and compares it to the
+locked hash — if `lock` and `check` walked different file sets, the hash would
+drift spuriously on every run. Unifying both on `core/discovery` (rather than
+`lock.ts` keeping its own markdown-only `walkMarkdown`) is what makes the parity
+hold; it also widened `lock`'s collection to include source-file entries (a
+lockfile generated before this change needs one `markspec lock` refresh).
+
+**One shared collector (#684).** The discover→parse walk is a single core
+function — `collectProjectEntries(root, io, { exclude })` in `core/collect/` —
+that `lock` (the pinner), `compile`, `fmt`, and `doctor` all call, so "the set
+of files whose edges are hashed" has exactly one definition and cannot drift
+between pinner and checkers. It relies on `discoverFiles`' default extension
+set, so no caller pins the extensions independently (a pinned value would
+re-open the divergence). `check` is the one deliberate exception: it needs each
+file's raw content and per-file directives (the fmt-drift `mdContents` and
+listing contexts), not just the flattened entries, so it keeps its
+`resolveScope`-based loop — equivalent by construction (same discovery, same
+`exclude:`, same default extensions). Before #684 the collector lived as a
+private helper in `cli/commands/lock.ts` that `compile`/`fmt` reached into and
+`doctor` re-rolled by hand — three copies that had to stay byte-identical.
 
 **`doctor` shares the same comparison (#658).** The offline edge-hash comparison
 `check`'s `MSL-L212` gate performs is factored into one pure helper —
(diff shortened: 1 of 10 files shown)


stasson:
## `/review` — 3 finder angles (behaviour-preservation · cross-file/correctness · conventions)

**Overview.** Extracts the three hand-synced discover→parse loops (`lock`'s private `collectEntries` that `compile`/`fmt` reached into, plus `doctor`'s #682 copy) into one exported `collectProjectEntries` in `core/collect/`, routing all four commands through it. **Cross-file: CLEAN** (exports resolve, no dangling imports, no cycle, no test breakage). **Conventions: CLEAN** (no `Deno.*` in the new core module, library-boundary imports, docs present).

Two findings from the behaviour-preservation angle:

### 1. Unreadable-file: throw → skip — **accepted** (documented)

The old `lock` collector read with `Deno.readTextFile` (throws on an unreadable discovered file); the shared collector uses the injected `io.readFile` (returns `undefined` → skips). This unifies all four callers on skip-on-unreadable — `doctor` already behaved this way. The trigger (a file yielded by `discoverFiles`, then unreadable at read time — a permission edge or a walk→read TOCTOU race) is a degenerate error state; skip-and-continue is the more robust default for a health-check/pin command than a crash. Documented in the collector's doc comment (`Unreadable files are skipped`). No change.

### 2. `compile --frozen` ignored `exclude:` — **fixed in this PR** (+ new e2e)

Pre-existing, not introduced here, but it directly contradicts the PR's parity thesis: `compile --frozen` collected with **no** `exclude`, so a project excluding an entry-bearing path — like this repo's own `skills/` — would count edges `markspec lock` never pinned, spuriously drifting the frozen check (`MSL-L212`, exit 1). Fixed by passing `project.yaml` `exclude:` (already loaded two lines above) to the collector, so the pinner and the `--frozen` checker use the same file set. Guarded by a new e2e (`compile_frozen_exclude_test.ts`) that was RED before the fix (spurious `MSL-L212`) and GREEN after.

Full suite after both: **3507 passed, 0 failed.**

Issue #645, opened 2026-07-01
MCP: discover a project in a subdirectory of the launch/workspace root


stasson:
Follow-up from #641 (deferred by design — spec decision D7).

## Context

The #641 fix resolves the MCP server's project root from an ordered candidate
list: `--root` > `MARKSPEC_PROJECT_ROOT` > `CLAUDE_PROJECT_DIR` > launch `cwd`.
Each candidate is a **start dir** from which discovery walks **upward** for
`project.yaml` / `.markspec.yaml`.

## Gap

When the editor opens a **monorepo at an ancestor** of the MarkSpec project
(e.g. workspace root `…/sdv`, project at
`…/sdv/CCS/HMI/Modules/Specification/cockpit-experience`), every automatic
candidate (`CLAUDE_PROJECT_DIR`, `cwd`) is an **ancestor**, so the upward walk
never reaches the project. Today's workaround is the explicit
`MARKSPEC_PROJECT_ROOT` / `--root` override pointing straight at the subdir.

## Why deferred

A bounded **downward** search from each candidate is ambiguous when a parent
holds multiple MarkSpec projects and needs a tie-break rule. Out of scope for the
#641 minimal fix (YAGNI), tracked here.

## Possible approaches

- Bounded-depth downward scan for `.markspec.yaml` / `project.yaml`, with a
  deterministic tie-break (nearest, or error on ambiguity).
- Multi-project support (server exposes / selects among several roots).

Effort: S. Depends on #641 landing first.

PR #691, opened 2026-07-03
fix(mcp): fall back to .markspec.yaml directory when no project.yaml exists

### LINKED ISSUE #647: MCP: .markspec.yaml-only project detects but leaves projectRoot undefined → getCompiled throws
Surfaced during the #641 review (pre-existing, not introduced by that change).

## Problem

In the MCP server's project resolution (`packages/markspec/mcp/project.ts`,
`createProject`), a discovery candidate is adopted when `detectMarkspecProject`
returns true — which accepts **either** `project.yaml` **or** `.markspec.yaml`.
But `projectRoot` is set from `discoverProjectRoot`, which keys **only** on
`project.yaml`.

So for a project that has a `.markspec.yaml` but **no** `project.yaml` anywhere
above the candidate:

- `markspecDetected = true` → the soft gate does **not** fire, and
- `projectRoot = undefined` → `getCompiled()` / `forceRefresh()` throw
  `"MarkSpec MCP server: no project.yaml found."`

The agent sees tools fail with an internal error instead of either working or
getting the actionable soft-gate message.

`.markspec.yaml`-only projects are valid per ADR-008 (the activation chain does
not require `project.yaml`), so this is a real (if narrow) gap.

## Scope note

This predates #641 — the pre-#641 code called the same two functions
back-to-back on `cwd`; the #641 resolution loop faithfully preserved the
semantic. It is **not** the #641 reporter's case (their project has both files).

## Possible fix

Give `.markspec.yaml`-only projects a coherent story: either treat the
`.markspec.yaml` directory as the `projectRoot` when no `project.yaml` exists, or
make the soft gate fire (with a message explaining a `project.yaml` is required
for compile-backed tools). Decide which is correct against ADR-008 / the config
loader's expectations.

Effort: S.

stasson:
## Summary

- Closes #647.
- A `.markspec.yaml`-only project (valid activation per ADR-008, no `project.yaml` anywhere upward) previously left `projectRoot` `undefined` even though `markspecDetected` was `true`, so the soft gate never fired and `getCompiled()`/`forceRefresh()` threw an internal `"no project.yaml found"` error instead of working.
- `createProject` now falls back to `discoverMarkspecRoot`'s directory when `discoverProjectRoot` finds nothing upward, so compile-backed tools (`entry_search`, `validate`, etc.) work fully for `.markspec.yaml`-only projects. `Project.config` stays `undefined` in this case since there's no `project.yaml` to load — confirmed via grep that no MCP tool/resource actually consumes `project.config` today.
- Updated ADR-028's Consequences section (which explicitly tracked this as follow-up #647) to reflect the fix.

## Test plan

- [x] Added a red→green unit test pair in `project_test.ts` verifying `projectRoot` falls back correctly and `getCompiled()` succeeds for a `.markspec.yaml`-only project; confirmed both fail without the `project.ts` fix.
- [x] `just build` (lint + test + typecheck + compile) passes.
- [x] Pre-push hook's full `just check` passed.

### DIFF
--- a/docs/architecture/adr-028-mcp-project-discovery.md
+++ b/docs/architecture/adr-028-mcp-project-discovery.md
@@ -133,10 +133,11 @@ multiple projects. Tracked as follow-up #645.
 - `Project` remains a plain, set-once, synchronously-constructed value — no new
   async re-resolution path, no new mutable state for tools/resources to reason
   about.
-- A `.markspec.yaml`-only candidate still detects a project
-  (`markspecDetected = true`) but can leave `projectRoot` undefined, exactly as
-  before this ADR (`discoverProjectRoot` only recognizes `project.yaml`).
-  Tracked as follow-up #647.
+- A `.markspec.yaml`-only candidate (`markspecDetected = true`, no
+  `project.yaml` anywhere upward) resolves `projectRoot` to the directory
+  containing `.markspec.yaml` (#647, fixed). Compile-backed tools work fully for
+  these projects; `Project.config` stays `undefined` since no `project.yaml`
+  exists to load.
 
 ## Alternatives considered
 
--- a/packages/markspec/mcp/project.ts
+++ b/packages/markspec/mcp/project.ts
@@ -15,6 +15,7 @@ import {
   compile,
   type CompileResult,
   type DeliveredDocument,
+  discoverMarkspecRoot,
   discoverProjectRoot,
   type EffectiveProfile,
   loadConfig,
@@ -202,7 +203,13 @@ export type InvalidationHandler = (
 
 /** Project context handle returned by {@linkcode createProject}. */
 export interface Project {
-  /** Discovered project root, or undefined when no `project.yaml` was found. */
+  /**
+   * Discovered project root, or undefined when neither `project.yaml` nor
+   * `.markspec.yaml` was found. Prefers the directory containing
+   * `project.yaml`; when only `.markspec.yaml` is found (a project
+   * activated per ADR-008 without a `project.yaml`), falls back to that
+   * directory (#647) so compile-backed tools still work.
+   */
   readonly projectRoot: string | undefined;
   /**
    * `true` when the workspace contains either `project.yaml` or
@@ -311,7 +318,12 @@ export async function createProject(env: ProjectEnv): Promise<Project> {
   for (const candidate of candidates) {
     if (await detectMarkspecProject(candidate, env.readFile)) {
       markspecDetected = true;
-      projectRoot = await discoverProjectRoot(candidate, env.readFile);
+      // A `.markspec.yaml`-only activation (no `project.yaml` anywhere
+      // upward) is valid per ADR-008; fall back to its directory so
+      // compile-backed tools still get a projectRoot instead of throwing
+      // "no project.yaml found" (#647).
+      projectRoot = await discoverProjectRoot(candidate, env.readFile) ??
+        await discoverMarkspecRoot(candidate, env.readFile);
       break;
     }
   }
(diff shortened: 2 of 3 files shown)


stasson:
## Code review

This PR fixes #647 by making `createProject` fall back to the directory containing `.markspec.yaml` when no `project.yaml` exists anywhere upward, so `.markspec.yaml`-only projects (valid per ADR-008) get a `projectRoot` instead of `markspecDetected=true` + `projectRoot=undefined` silently throwing inside `getCompiled`/`forceRefresh`. The change is small, well-tested (red→green unit tests confirmed), and `just build`/pre-push `just check` both pass clean.

I ran an 8-angle finder pass + independent verification on each surviving candidate. Nothing blocks merge — the fix is correct for the case it targets (the "recommended," co-located `.markspec.yaml` layout). Everything below is a non-blocking, edge-case observation surfaced while touching this code, worth follow-up issues rather than changes to this PR:

1. **`packages/markspec/mcp/project.ts:325-326`** — When an *unrelated ancestor* `project.yaml` exists further up the tree than a closer `.markspec.yaml` (e.g. a monorepo root `project.yaml` above a subproject activated only by its own `.markspec.yaml`), `discoverProjectRoot(candidate)` wins the `??` and resolves `projectRoot` to that distant, unrelated ancestor — the new `discoverMarkspecRoot` fallback never fires. **Verified pre-existing**: `discoverProjectRoot(candidate)` already had this exact precedence before this PR (the old line was `projectRoot = await discoverProjectRoot(candidate, env.readFile);` with no fallback at all), so this isn't a regression from this diff — just a latent quirk in the same lines this PR touches, worth a follow-up issue since it can silently defeat #647's intent in nested-activation monorepo layouts.

2. **`packages/markspec/lsp/server.ts:349`** (unchanged, for comparison) — LSP's analogous no-`project.yaml` fallback is `discoverProjectRoot(rootPath, readFile) ?? rootPath` (raw, unwalked editor root), while MCP's new fallback walks up to the actual `.markspec.yaml` directory via `discoverMarkspecRoot`. Since `readMarkspecYaml` reads `.markspec.yaml` at the *exact* given path (no upward walk), LSP's fallback will silently fail to load the profile whenever the editor's opened folder is a subdirectory of where `.markspec.yaml` actually lives — a case MCP now handles correctly. Worth a follow-up issue to have LSP prefer its already-computed `discoverMarkspecRoot` result (it's computed a few lines earlier for the event-log gate but not reused for `projectRoot`) rather than the raw `rootPath`.

3. **`packages/markspec/mcp/project.ts:354-365` (`discoverTracked`)** — The new fallback makes a previously-dead branch reachable: if `.markspec.yaml` isn't co-located with the project's tracked files (sits at a distant ancestor, or in a narrower subdirectory than the docs), `discoverTracked`'s walk — which has no gitignore/`exclude:` filtering — will now silently walk too broad a subtree (unrelated sibling projects, extra startup cost) or too narrow one (missing entries, no diagnostic), where before it just threw a clear, loud error. This requires a self-inflicted, uncommon layout (the profile module's own docstring says `.markspec.yaml` is meant to sit "next to `project.yaml`"), and is in the same accepted-limitation class as ADR-028 §7's monorepo blind spot (tracked as #645). Non-blocking; a fast-follow test/issue would be good documentation of the boundary.

4. **`packages/markspec/mcp/project.ts:318-329`** — Minor hygiene note: `detectMarkspecProject`, `discoverProjectRoot`, and `discoverMarkspecRoot` are three structurally identical upward walks; the winning candidate can be walked up to 3 times since `detectMarkspecProject`'s discovered location/file isn't threaded through to the two calls that follow. Confirmed real, but severity is negligible (once per MCP server startup, cheap local `readFile`s) — not worth its own fix, just a "nice to have if touching this file again."

Nice catches from the tests: the red→green pair (`createProject: falls back to .markspec.yaml directory...`, `getCompiled: succeeds for a .markspec.yaml-only project`) correctly exercises the case this PR is meant to fix.

stasson:
Filed follow-up issues for the three review findings above: #693 (ancestor project.yaml precedence bug), #694 (LSP fallback parity), #695 (scope-alignment risk in the .markspec.yaml fallback).

Issue #733, opened 2026-07-03
debt(core): make inline code spans structurally opaque to prose lint + MSL-M060


stasson:
Follow-up from #719 (epic #717). #719 locked the opacity of the uxil DSL surface (lowercase `ux:` refs, `$`-shape citations, unknown fenced blocks) with regression tests, but deliberately did **not** close the underlying structural gap: inline code spans are not exempt from prose analysis or the MSL-M060 modal scanner.

## Gap

Prose-lint rules (Q500 xref, EARS, modal, INCOSE, passive) read a paragraph's verbatim `content.text`, which includes inline-code-span text **and its backticks**. The MSL-M060 modal scanner in `core/parser/body_tokens.ts` is line-based and inline-code-blind. `collectVerbatimLines` covers only block-level `code`/`feature`/`math`, never inline spans. So a code span containing an uppercase modal keyword or a PascalCase term still draws a diagnostic:

- `` `SHALL` `` in prose → **MSL-M060** ("modal keyword 'SHALL' should be lowercase")
- `` `SHALL` and `HomeScreen` `` → **MSL-Q500** (xref-glossary-undefined; the flagged phrase even swallows the backticks)

The uxil DSL surface dodges this only because it is lowercase/symbolic and `$`-refs are silenced by the typl entity-ref index — so today's guarantee is held by grammar, not structurally.

## Proposal (Scope B, deferred from #719)

- Strip `code`/`inlineCode` from the prose-lint analyzed text so Q500 / EARS / modal / INCOSE / passive never see span innards — mirroring the existing block-level `code`/`feature`/`math` exclusion.
- Make the `body_tokens.ts` modal / EARS / entity scanners inline-code-aware so MSL-M060 does not fire on span content.
- Must **not** break typl `$`-ref Q500 silencing (`buildIdentifierIndex`) or typl's own inline declaration extraction (`extractTyplInlines`).

## Acceptance

- `` `SHALL` `` and a PascalCase term inside a code span draw no MSL-M060 / MSL-Q500.
- typl inline declarations + `$`-ref silencing regression-proven unchanged.
- The #719 regression suite (`tests/e2e/opaque_dsl_test.ts`) still passes.

Load-bearing (ADR-016 body-token AST, ADR-021 prose analysis) — medium regression risk; deliberately deferred out of #719's "smallest-first, no new behavior" scope.

Issue #717, opened 2026-07-03
uxil declaration DSL + typl namespacing, on shared declaration-surface machinery


stasson:
**Epic** · milestone `v0.11.0` · label `epic:uxil`

## Summary

Three related deliverables on one substrate, smallest-first:

- **§A — shared declaration-surface machinery:** generalize typl's declaration surfaces (add a **table surface** via `msTableCaption`) + a formal entry-local **base-resolution** rule set.
- **§B — uxil (UX Interaction Language):** a second declaration DSL on that machinery — typed UI/HMI surfaces + interactions, a `ux:` URI reference grammar, a corpus `uxRegistry`, a `UXIL-0xx` diagnostics family, and LSP. Sibling of typl (ADR-019 is the architectural template).
- **§C — typl improvements** on the same machinery: **namespacing** for published symbols, the table surface, and a **payload bridge** binding uxil event payloads to typl shapes.

## Motivation

Automotive HMI teams have a deployed UX identity convention (`screenId/elementId[:itemKey]` test tags shared by UI Automator / Compose / Espresso selectors and OTel-aligned analytics) but nothing *declares* which ids exist or what each element affords — so specs, tests, journeys, and telemetry cannot be validated against the UI surface. typl proved the pattern (prose-embedded typed declarations + registry + diagnostics) and has the same unsolved namespacing gap this epic closes.

## Stories

**Phase 0 — Tier 1 (unblock ACME now)**

- [x] #719 — S1 · guarantee `check` treats unknown code-span DSLs + fence tags as opaque

**Phase 1 — §A shared machinery (foundation)**

- [x] #720 — S2 · shared declaration-surface machinery (typl trio → common AST) *[byte-identical refactor]*
- [x] #721 — S3 · table declaration surface via `msTableCaption`
- [x] #722 — S4 · entry-local base-resolution engine

**Phase 2 — typl on the machinery (§C.1–C.2)**

- [x] #723 — S5 · typl published/namespaced tier (dotted paths, declared-once, retire TYPL-002/003) ⚠️ *compat gate + schema bump*
- [x] #724 — S6 · typl table surface

**Phase 3 — uxil DSL (§B)**

- [x] #725 — S7 · uxil parser (`ux:` URI reference grammar + declaration forms)
- [x] #726 — S8 · uxil compiler + `uxRegistry` + machine projection *[XL — the heart]*
- [x] #727 — S9 · `UXIL-0xx` diagnostics family
[...]

PR #734, opened 2026-07-03
test(core): guarantee check treats unknown code-span DSLs + fence tags as opaque

### LINKED ISSUE #719: story(core): guarantee check treats unknown code-span DSLs + fence tags as opaque
Part of #717 (Tier 1 — unblock ACME now). Smallest-first; guarantee + regression coverage, no new behavior.

## Goal
Guarantee `markspec check` treats unknown DSL content as opaque so the `@acme/hmi` profile can host the uxil vocabularies via an external linter before uxil lands in core (§717 Staging Tier 1):
- inline **code spans** with unknown declaration/citation syntax (e.g. `` `ux:media.home/play` ``) MUST NOT produce diagnostics and MUST be exempt from prose lint;
- fenced blocks with an unknown info-string/tag MUST pass through untouched (no parse, no format rewrite, no diagnostic).

## Acceptance
- e2e: a doc with `ux:`-style code spans + an unknown fenced tag passes `markspec check` clean (no MSL-*, no Q500).
- Locks the behavior with a regression test so future changes can't silently start flagging unknown code-span DSLs.
- Documented as a stability guarantee.

stasson:
Closes #719.

Part of epic #717 (Staging Tier 1). Locks the stability guarantee that `markspec check` treats unknown DSL content as opaque, so the `@ampere/seed` profile can host the uxil vocabularies via an external linter before uxil lands in core.

## What

Scope A — lock + document, **no production change**. The uxil DSL surface — lowercase `ux:` refs, leading-slash/dot element forms, `$`-prefixed typl shape citations, and fenced blocks with an unknown info-string — is already inert to every active `check` / `lint` / `fmt` rule. This PR pins that with regression coverage + a documented guarantee.

## Evidence (why no production change)

A representative doc (all four uxil declaration forms + a `ux:` citation + the scheme-less wire form + a camelCase tag + a `$`-shape citation + an unknown `uxil` fence):

- `markspec check` → exit 0, no MSL-*, no Q500
- `markspec lint` → zero diagnostics
- `markspec fmt` → byte-identical (spans + fence untouched)
- `markspec compile` → `entry.types` absent (`ux:` not parsed as typl)

## Changes

- `tests/e2e/opaque_dsl_test.ts` — the named #719 regression suite (check / lint / fmt / compile).
- `packages/markspec/core/parser/body_tokens_test.ts` — unit: an unknown info-string fence is verbatim (no tokens), like a known one.
- `docs/spec/language/language.md` — "Opaque DSL content" note.
- `docs/spec/internal/markspec-prose-analysis.md` §1.2 — stability guarantee, scoped honestly.

## Scope boundary (honest)

The guarantee covers the uxil DSL surface, **not** arbitrary text. Inline code spans are not yet _structurally_ excluded from prose lint / the MSL-M060 modal scanner, so an uppercase modal keyword or a PascalCase term inside a span is still analyzed. Empirically `` `SHALL` `` fires MSL-M060 and `` `SHALL` and `HomeScreen` `` fires MSL-Q500. Closing that structural gap is deferred follow-up #733 (Scope B) — deliberately out of this story's "smallest-first, no new behavior" scope.

## Verification

`just check` (3535 passed, 0 failed), `deno fmt --check`, `dprint check`, `just compile` — all green.

### DIFF
--- a/docs/spec/internal/markspec-prose-analysis.md
+++ b/docs/spec/internal/markspec-prose-analysis.md
@@ -103,6 +103,20 @@ further by default:
 `prose.scope.blocks` (§4.1) opts the opt-in blocks in. The never-row is not
 profile-overridable (it would contradict core-data-model §2.5).
 
+**Unknown DSL content is opaque (stability guarantee, #719).** A fenced block is
+verbatim regardless of its info-string: an unknown language tag (`uxil`) is
+skipped exactly like `rust`, so no rule sees inside it. The uxil
+declaration/citation surface embedded in inline code spans — lowercase `ux:`
+refs, leading-slash / dot element forms, and `$`-prefixed shape citations — is
+likewise inert to every active rule, which lets the `@ampere/seed` profile host
+those vocabularies via an external linter before uxil lands in core (epic #717,
+Tier 1). The guarantee is scoped to that DSL surface, **not** to arbitrary text:
+inline code spans are not yet _structurally_ excluded (rules read a paragraph's
+verbatim `content.text`, which includes the span), so an uppercase modal keyword
+or a PascalCase term inside a span is still analyzed. Closing that structural
+gap is deferred follow-up #733. Regression-locked by
+`tests/e2e/opaque_dsl_test.ts`.
+
 ### 1.3 Intent and non-goals
 
 - **Inline markers are signal.** `ModalMarker` and `EntityRef` (core-data-model
--- a/docs/spec/language/language.md
+++ b/docs/spec/language/language.md
@@ -214,11 +214,11 @@ fn debounce_input(raw: u16, window_ms: u32) -> u16 {
 
 **Opaque DSL content.** A fenced block whose language identifier is unknown to
 the toolchain is passed through verbatim — never parsed, reformatted, or
-prose-analyzed. Inline code spans carrying an unknown declaration or citation
-DSL — for example the interaction-reference form `ux:media.home/play` — are
-likewise left untouched by `markspec fmt` and `markspec check`. External
-profiles rely on this to host their own embedded vocabularies before those
-vocabularies land in core.
+prose-analyzed. `markspec fmt` also leaves inline code spans byte-for-byte
+unchanged, and an unknown reference DSL embedded in a span — for example the
+interaction-reference form `ux:media.home/play` — draws no `markspec check`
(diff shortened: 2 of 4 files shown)


stasson:
## `/review` — self-review (3 finder angles, medium effort)

Reviewed across test-correctness, doc-accuracy-vs-code, and reuse/altitude angles. Every factual claim in the docs was verified against the code. Two actionable findings surfaced; **both fixed in `f7dca3d`**.

### Fixed

1. **`tests/e2e/opaque_dsl_test.ts` — `fmt` test over-asserted.** It compared whole-document byte equality (`after === UX_DOC`), coupling the DSL-opacity guarantee to the surrounding prose wrap. `markspec fmt` reflows entry-body prose through dprint, so an orthogonal embedded-dprint bump could rewrap the EARS sentence and raise a false "opaque DSL broken" signal while the DSL stayed verbatim. Scoped to the actual guarantee: the `uxil` fence block + every `ux:` span survive fmt byte-for-byte.
2. **`docs/spec/language/language.md` — public note over-claimed.** The "Opaque DSL content" note used "likewise", implying inline code spans get the same *structural* prose-analysis exemption as fenced blocks. They don't — prose rules read the paragraph's verbatim `content.text` (which includes the span); the `ux:` forms are inert by grammar, not structurally. Reworded to match the honest internal note and avoid misleading profile authors.

### Verified correct — no change

- `compile` test's `entry.types === undefined` has real teeth (inline typl also populates `entry.types`, so a regression would fail it).
- Permissions sufficient under the helper's default `--allow-read --allow-write`.
- `markspecPersist` temp-dir cleanup correct (try/finally).
- body_tokens line filter accurate; docs claims (fence verbatim regardless of info-string; spans read via `content.text`; bare `$`-citations inert) all confirmed against the code.

### Acknowledged, kept (non-blocking)

- New `body_tokens` unit test overlaps the existing `rust`-fence test's code path (there is no known-vs-unknown-language branch to regress independently); kept as the named #719 anchor + the `$Ref`-in-fence assertion.
- Local `BASE_FILES` scaffold duplicates `check_project_test`'s exported one (diverges only by the `markspec-schema` pin the strict "no MSL-*" assertion needs); inlining is the established e2e pattern.

> Note: the review fixes are a follow-up commit (`f7dca3d`) because the auto-mode guard blocks amending an already-pushed commit — please **squash-merge** to keep one commit per PR.

PR #737, opened 2026-07-03
feat(core): entry-local base-resolution engine for declaration surfaces

### LINKED ISSUE #722: story(core): entry-local base-resolution engine for declaration surfaces
Part of #717 §A. Depends on: S2 (shared machinery). The normative rule set for relative-ref resolution.

## Goal
Lexical, **entry-local** base-resolution for declaration surfaces:

1. Only *declarations* create bases; *citations* never do. Declarations are recognized syntactically and legal only in the DSL's declaring entry type.
2. Bases are entry-local — no document-ambient scope (an entry means the same thing quoted standalone).
3. A relative ref resolves by walking **up the AST**, innermost base wins: enclosing table caption base → enclosing declaration-bullet subtree base → entry root declaration → else diagnostic.
4. Scope is **structural** (list nesting / table membership), never sequential — reordering never rebinds a ref; siblings don't affect each other.
5. Exactly **one root declaration** per declaring entry (diagnostic on zero or several).
6. Declarations/citations sit in **code spans** (backticks delimit only, never part of the token).

```
resolve(ref, node):
  if absolute(ref): return ref
  for anc in ancestors(node):
    if base(anc): return join(base(anc), ref)
  error "relative ref with no base in scope"
```

## Acceptance
- `resolve()` unit-tested per rule (innermost-wins, no-base error, structural-not-sequential, one-root invariant).
- Diagnostics: "relative ref with no base in scope"; zero/multiple root declarations.
- DSL-agnostic (typl + uxil both consume it).

stasson:
Closes #722.

uxil epic (#717 §A). The DSL-agnostic **normative rule set** for resolving a relative reference against the base an enclosing declaration establishes. typl's published tier (S5) and uxil (S7) both consume it; the engine carries no vocabulary.

## API (`core/decl/resolve.ts`)

- `resolveRef(ref, scope, ops)` — an absolute ref passes through unchanged; a relative ref joins the base of the **nearest enclosing scope** that carries one (innermost wins, single join — bases are not accumulated up the chain); no base in any ancestor → `{ ok: false, reason: "no-base-in-scope" }`.
- `checkSingleRoot(roots)` — the one-root invariant; else `{ ok: false, reason: "no-root" | "multiple-roots", roots }`.
- Scope is an abstract `BaseScope { base?, parent? }` chain; ref syntax is injected via `RefOps { isAbsolute, join }`. So the engine is vocabulary-agnostic and needs **neither** the table surface (S3) **nor** a DSL consumer to land.

## The six rules → where each lives

1. Only declarations create bases — the caller sets `BaseScope.base` only for declaration scopes.
2. Bases are entry-local — the engine only walks the `parent` chain it's handed; no global lookup.
3. Innermost base wins — `resolveRef` returns on the first ancestor with a base.
4. **Structural, not sequential** — the engine follows `parent` links only, never siblings, so reordering declarations can't rebind a ref. (Directly tested.)
5. Exactly one root per entry — `checkSingleRoot`.
6. Declarations sit in code spans — a surface concern (S2 + the DSL recognizer); the engine sees only the extracted ref token.

## Diagnostics

The engine returns **typed reasons** (`no-base-in-scope`, `no-root`, `multiple-roots`) that map 1:1 to the three diagnostics in the issue. Messages + codes stay DSL-owned (typl TYPL-0xx, uxil UXIL-0xx in S9) — matching the existing typl diagnostics model, keeping the engine DSL-agnostic.

## Scope boundary

Engine + per-rule tests only (9 tests: innermost-wins, single-join-not-cumulative, base-less-scope skip, no-base error, structural-not-sequential + order-independence, absolute passthrough, one-root ×3). Building `BaseScope` chains from real surfaces — including the table-caption base ("caption base surfaced to the resolver (S4)") — is **S3** and the DSL consumers (**S5/S7**). `core/decl/` stays internal (no `core/mod.ts` re-export); its consumers live inside `core/`.

## Verification

`just check` **3564 passed / 0 failed**, `deno fmt --check`, `dprint check`, `just compile` — all green.

### DIFF
--- a/packages/markspec/core/decl/mod.ts
+++ b/packages/markspec/core/decl/mod.ts
@@ -4,8 +4,10 @@
  * Shared declaration-surface machinery (uxil epic #717 §A). DSL-agnostic
  * walkers that extract declaration sites from an entry body's three
  * embedding surfaces — fenced code blocks, bullet paragraphs, and inline
- * code spans — parameterized by a recognizer. typl consumes these today;
- * uxil (§B) and the table surface (S3) ride on the same substrate.
+ * code spans — parameterized by a recognizer, plus the entry-local
+ * base-resolution engine (S4) that resolves relative refs against the base
+ * an enclosing declaration establishes. typl consumes these today; uxil
+ * (§B) and the table surface (S3) ride on the same substrate.
  */
 
 export type {
@@ -20,3 +22,6 @@ export {
   extractInlineDeclarations,
   stripCodeSpanDelimiters,
 } from "./surfaces.ts";
+
+export type { BaseScope, RefOps, RefResolution, RootCheck } from "./resolve.ts";
+export { checkSingleRoot, resolveRef } from "./resolve.ts";
--- a/packages/markspec/core/decl/resolve.ts
+++ b/packages/markspec/core/decl/resolve.ts
@@ -1,0 +1,121 @@
+/**
+ * @module core/decl/resolve
+ *
+ * Entry-local base-resolution engine (uxil epic #717 §A). The DSL-agnostic
+ * *normative rule set* for resolving a relative reference against the base
+ * an enclosing declaration established. typl's published tier and uxil both
+ * consume it; the engine knows nothing about either vocabulary.
+ *
+ * The six rules from #722, and where each lives:
+ *
+ *   1. Only *declarations* create bases; *citations* never do. — The caller
+ *      sets {@linkcode BaseScope.base} only for declaration scopes; a scope
+ *      built from a citation leaves it `undefined`.
+ *   2. Bases are entry-local (no document-ambient scope). — The engine only
+ *      ever walks the `parent` chain it is given; there is no global lookup.
+ *   3. Innermost base wins. — {@linkcode resolveRef} returns on the first
+ *      ancestor (nearest first) that carries a base.
+ *   4. Scope is structural, never sequential. — The engine follows `parent`
+ *      links only; siblings are unreachable, so reordering declarations can
+ *      never rebind a ref.
+ *   5. Exactly one root declaration per declaring entry. —
+ *      {@linkcode checkSingleRoot}.
+ *   6. Declarations/citations sit in code spans. — A surface concern
+ *      (core/decl/surfaces + the DSL recognizer); the engine sees only the
+ *      already-extracted ref token.
+ *
+ * The caller builds the {@linkcode BaseScope} chain from the entry's
+ * structural nesting (list membership, table membership) — that mapping is
+ * DSL- and surface-specific and lands with the table surface (S3) and the
+ * DSL consumers (S5/S7).
+ */
+
+/**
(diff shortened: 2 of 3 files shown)


stasson:
## `/review` — self-review (2 finder angles: correctness/rule-coverage + API/conventions)

**Verdict: engine is correct and faithfully implements all six normative rules.** Both finders independently flagged one vacuous test tail; fixed in `838b289`. Everything else clean.

### Correctness — engine correct, all 6 rules mapped

- **Innermost-wins / single-join** — the loop returns on the first ancestor with a base and joins exactly once; no accumulation possible (return is inside the `if`). Proven by the "single join, not cumulative" and base-less-scope-skip tests.
- **`base !== undefined` (not truthiness)** — confirmed correct: an explicitly empty-string base is a *present* base and must win innermost, not fall through. (Now pinned by a new test — see below.)
- **Rules 1/2** — structural guarantees (caller sets base only for declarations; engine walks only the injected `parent` chain, no global lookup). **Rule 5** — one/zero/multiple all covered. **Rule 6** — surface concern, correctly out of scope.
- **API** — `RefResolution` / `RootCheck<T>` discriminated unions complete; `checkSingleRoot` returning the full `roots` set is exactly what a `multiple-roots` diagnostic needs; generic `T` lets a host recover its own root objects. Nothing missing.

### API / conventions — clean

- **Boundary** — `resolve.ts` is pure (zero imports), re-exported from `core/decl/mod.ts`, absent from `core/mod.ts` — the same pattern as sibling `surfaces.ts` (core-internal, typl/uxil consume via `../decl/mod.ts`). No cycle, no upward import.
- **Naming** — no collisions in `core/`; the only proximity (`resolveRef` vs `core/lock`'s `resolveReferences`) is a different domain, distinct name.
- **Message-deferral** — returning typed reasons (`no-base-in-scope`/`no-root`/`multiple-roots`) instead of English strings is the correct call for a DSL-agnostic engine: typl and uxil own separate diagnostic namespaces + wording (rule 6 frames presentation as a surface concern).

### Fixed (`838b289`)

- **Vacuous rule-4 tail** (both finders) — the old `_siblingC` binding was never reachable through `parent` links and its assertion duplicated the prior one, so it couldn't catch a sibling-traversal regression. Rewrote to three siblings where A and B each declare their own base and C declares none, genuinely proving a sibling's base is invisible to its siblings.
- **Empty-base test added** — pins that `base !== undefined` (not `if (s.base)`) is intentional, so a later truthiness "simplification" is caught.

### Considered, not changed

- **Cycle guard** (low, one finder) — a cyclic `parent` chain would spin forever. Skipped: callers build acyclic `BaseScope` trees from acyclic ASTs, and the sibling `surfaces.ts` walkers recurse AST containment with the same unguarded acyclic assumption. Adding a guard here only would be inconsistent over-engineering for an impossible-in-practice caller bug.

> Note: like #734/#736, the review fix is a follow-up commit (auto-mode guard blocks amending a pushed commit) — please **squash-merge** to keep one commit per PR.

PR #738, opened 2026-07-03
feat(core): table declaration surface via msTableCaption

### LINKED ISSUE #721: story(core): table declaration surface via msTableCaption
Part of #717 §A. Depends on: S2 (shared machinery).

## Goal
Add a **table declaration surface** on the shared machinery: each table row is one declaration; the `Table:` caption (`msTableCaption`) may carry a **base** (a ref in the caption) consumed by base-resolution (S4).

## Acceptance
- A table under a declaring entry parses each row to the common declaration AST.
- Caption-carried base surfaced to the resolver (S4).
- Unit tests: rows→declarations, caption base extraction, malformed rows.

stasson:
Closes #721.

uxil epic (#717 §A). A fourth embedding surface on the shared declaration machinery (alongside fence, bullet, inline): each **data row** of a GFM table is one declaration, and the table's `Table:` caption is surfaced as the base candidate the entry-local resolver (S4) consumes.

## API (`core/decl/table.ts`)

- `extractTableDeclarations(blocks, rowToSource)` → `TableRowDeclaration[]`, each `{ source, range, captionText? }`. Walks the body AST (recursing into list items); for each data row, `rowToSource(cells)` reduces the row to a declaration source, or returns `undefined` to skip it (**the malformed-row case**).
- `RowRecognizer = (cells: string[]) => string | undefined` — the only DSL-specific input.

## Two AST realities that shaped the design

1. **Rows carry no individual range** — a `TableNode` has one range for the whole table and its verbatim `raw`; cells are stored as text only. So a row's range is **line-precise, derived**: `table.range.start.line + 2 + rowIndex` (past the GFM header + delimiter), with the end column spanning the row's `raw` line. Diagnostics land on the offending row, not the whole table.
2. **Captions aren't linked to their table** (`CaptionNode.block` is never populated). So a table is paired with its `Table:` caption by **adjacency + the caption's `above`/`below` position** — matching `assignCaptionPositions` in `core/ast/build.ts` (position `above` ⇒ the caption precedes its table; `below` ⇒ follows it). The raw caption text rides on each row as `captionText`; the DSL parses the actual base ref out of it (keeping the ref grammar DSL-side, like S4's injected `RefOps`).

## Tests

Colocated `table_test.ts` (8 tests) with a **non-typl** toy recognizer (`@`-prefixed rows) proving DSL-agnosticism: rows→declarations, non-matching rows skipped, line-precise per-row range, caption above / below / non-Table-ignored / absent, and nested-in-list recursion.

## Scope & merge note

Independent of **S4 (#737)** — needs only S2's AST types (parallel siblings of S2). Both #721 and #737 append exports to `core/decl/mod.ts`, so **whichever merges second needs a trivial rebase** of the export list (no logic conflict).

## Verification

`just check` **3563 passed / 0 failed**, `deno fmt --check`, `dprint check`, `just compile` — all green.

### DIFF
--- a/packages/markspec/core/decl/mod.ts
+++ b/packages/markspec/core/decl/mod.ts
@@ -3,9 +3,9 @@
  *
  * Shared declaration-surface machinery (uxil epic #717 §A). DSL-agnostic
  * walkers that extract declaration sites from an entry body's three
- * embedding surfaces — fenced code blocks, bullet paragraphs, and inline
- * code spans — parameterized by a recognizer. typl consumes these today;
- * uxil (§B) and the table surface (S3) ride on the same substrate.
+ * embedding surfaces — fenced code blocks, bullet paragraphs, inline code
+ * spans, and table rows — parameterized by a recognizer. typl consumes
+ * these today; uxil (§B) rides on the same substrate.
  */
 
 export type {
@@ -20,3 +20,6 @@ export {
   extractInlineDeclarations,
   stripCodeSpanDelimiters,
 } from "./surfaces.ts";
+
+export type { RowRecognizer, TableRowDeclaration } from "./table.ts";
+export { extractTableDeclarations } from "./table.ts";
--- a/packages/markspec/core/decl/mod.ts
+++ b/packages/markspec/core/decl/mod.ts
@@ -2,7 +2,7 @@
  * @module core/decl
  *
  * Shared declaration-surface machinery (uxil epic #717 §A). DSL-agnostic
- * walkers that extract declaration sites from an entry body's three
+ * walkers that extract declaration sites from an entry body's four
  * embedding surfaces — fenced code blocks, bullet paragraphs, inline code
  * spans, and table rows — parameterized by a recognizer. typl consumes
  * these today; uxil (§B) rides on the same substrate.
--- a/packages/markspec/core/decl/table.ts
+++ b/packages/markspec/core/decl/table.ts
@@ -26,19 +26,17 @@ import type {
   SourceRange,
   TableNode,
 } from "../ast/nodes.ts";
+import type { BlockDeclaration } from "./surfaces.ts";
 
 /**
- * A declaration found on one data row of a table. `source` is the row
- * reduced to a declaration by the DSL's recognizer; `range` is the row's
- * line-precise span; `captionText` is the enclosing table's `Table:`
- * caption text, when the table has an adjacent one — the DSL parses a base
- * ref out of it and hands that to the resolver (S4).
+ * A declaration found on one data row of a table. Extends the shared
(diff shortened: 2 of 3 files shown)


stasson:
## `/review` — self-review (2 finder angles: correctness + API/conventions)

Both finders confirmed **top-level table handling is fully correct** (per-row line/column math, caption pairing, ragged/zero-row/ordering). They surfaced two real correctness bugs isolated to **list-nested tables**, plus three conventions items — all fixed in `be3cf48`.

### Fixed — correctness (nested tables)

1. **Caption pairing trusted `CaptionNode.position`, which is unreliable in list items.** The builder runs `assignCaptionPositions` only on top-level blocks, so a caption inside a list item keeps the default `"below"`. The old code therefore *dropped* an above-the-table caption in a list, and attributed a between-tables caption to the wrong table when nested. `adjacentTableCaption` now **re-derives above/below from local adjacency + captionability** (mirroring `assignCaptionPositions`), correct at top level and nested alike — and no longer depends on the stored position at all.
2. **Per-row range columns were off by the indent for nested tables.** `verbatimSlice` de-indents `raw`, so a hardcoded `column: 1` was short by the indent width. Columns now track `table.range.start.column` (`column .. column + rowLen`); top-level (column 1) is unchanged, nested is now correct. Line numbers were always correct.

### Fixed — conventions

3. **`TableRowDeclaration` now `extends BlockDeclaration`** instead of re-declaring `source`/`range`, matching the module's compose-on-`BlockDeclaration` convention (typl surfaces alias it) — so the four surfaces stay in lockstep if the shared shape changes.
4. **Barrel doc**: "three" → "four" embedding surfaces.

### Fixed — test gaps

5. Added tests: a nested caption (default `"below"` position) still attaches; a caption **between two tables** captions the following one (cross-attachment lock); indented-table row columns track the table column. **11 tests total.**

### Verified clean — no change

- Dependency-flow / boundary (imports `../ast` + `./surfaces` only; re-exported from `core/decl/mod.ts`, not `core/mod.ts`), naming parity, `readonly` usage.
- The conditional-spread `...(captionText !== undefined ? {…} : {})` (omit-undefined) is an established in-repo convention (`exactOptionalPropertyTypes` is off but the discipline is used across the AST layer). `RowRecognizer: cells => string | undefined` (recognize-and-project in one call) is the right signature since a row's source must be projected from cells.

## Verification

`just check` green, `deno fmt --check`, `dprint check`, `just compile` — all pass.

> Note: review fix is a follow-up commit (auto-mode guard blocks amending a pushed commit) — please **squash-merge**. Reminder: #721 and #737 both touch `core/decl/mod.ts`, so the second to merge needs a trivial export-list rebase.

Issue #699, opened 2026-07-03
SECURITY: profile-delivered document can symlink-exfiltrate arbitrary local files via MCP (#676)


stasson:
## Summary

A profile-delivered document path (ADR-030) is validated only as a **string** (no `..`, no absolute) — nothing canonicalises the resolved filesystem target before reading it. A malicious profile can deliver a `.md` path that is actually a **symlink to an arbitrary local file** (e.g. `~/.ssh/id_rsa`, `~/.aws/credentials`), and the MCP server will return that file's raw bytes to any client. This is a silent, zero-interaction local-file exfiltration vector.

## Vector

- `extends: git+https://…/evil-profile.git#v1` — git preserves symlinks on checkout; nothing in `core/profile/git-cache.ts` strips them.
- The profile declares `delivers: [{ path: "ref/arch.md", corpus: true }]` where `ref/arch.md` is a symlink to `~/.ssh/id_rsa`. The `.md`-extension check (`PROFILE-DELIVERS-004`) validates the declared **path string**, not the symlink target.
- MCP `markspec://delivered/{profileId}/{path}` returns raw bytes for **every** delivered document — the `listResourceDescriptors` loop (`packages/markspec/mcp/resources/mod.ts`) does not filter on `doc.corpus`, and `readDeliveredDocument` (`packages/markspec/mcp/project.ts`) does a bare `env.readFile(doc.absPath)`.
- An AI coding agent doing routine `resources/list`/`resources/read` (which this project's own MCP design assumes happens automatically, ADR-023) receives the victim's private key byte-for-byte in its context.

## Where

- `packages/markspec/core/profile/manifest.ts` (string-only path check)
- `packages/markspec/core/profile/delivered.ts` (`loadDeliveredCorpus` bare read)
- `packages/markspec/mcp/resources/mod.ts`, `packages/markspec/mcp/project.ts` (`readDeliveredDocument`)

## Fix direction

Before reading a delivered document (both `loadDeliveredCorpus` and `readDeliveredDocument`), `lstat`/`realpath` the resolved `absPath` and reject (a `PROFILE-DELIVERS-005`-style diagnostic) if it is a symlink, or if its canonicalised path falls outside the profile's `baseDir`. Consider whether docs-only (non-corpus) delivered documents should be MCP-readable at all.

Introduced with #676 (`feat(core): profile-delivered documents`, ADR-030).

_Found in post-merge review of PRs merged 2026-06-26 → 07-03. Filed as a public issue per maintainer decision._

PR #740, opened 2026-07-04
feat(core): federated upstream slice 1 — origin union + snapshot hydration

stasson:
Slice 1 of the federated-upstream-resolution design (docs/wip/2026-07-04-federated-upstream-resolution-design.md, carried on this branch): the pure-core foundation for cross-repo traceability references. No behavior is wired into CLI/LSP/MCP yet — that is slice 4; fetchers and lockfile integration are slices 2-3.

## What this adds (all pure core, no I/O of its own)

- `EntryOrigin` becomes the discriminated union its doc comment anticipated: `profile` (ADR-030, unchanged) + new `upstream` kind; `formatEntryOrigin` switches on kind; new `sameOriginSource` helper replaces the two unnarrowed `.profileId` comparisons in the corpus collision pass.
- `deserializeEntry` — exact inverse of `serializeEntry` (wire round-trip is whole-entry deep equality; `type`/`id` field presence restored to parser-canonical shape).
- `checkSnapshotSchema` + `extractSerializedEntries` — snapshot reading for both compile-output forms (inline `compiled.json` and NDJSON), with `UPSTREAM-SNAPSHOT-001` (schema skew, rejected before any data-file read — pinned by a recording-reader test), `-002` (missing file), `-003` (malformed JSON / non-object entry values / path escape).
- `loadUpstreamCorpus` — hydrates cached upstream snapshots into read-only `Entry[]` stamped `origin: { kind: "upstream", upstreamId, version }`, with per-upstream failure isolation and the authoritative-source rule (already-origin-carrying snapshot entries are re-exports and are skipped — what makes the diamond/root-aggregation topology collision-free).

## Hardening (from the whole-branch review)

- Valid-JSON-but-non-object entry values (e.g. a `null` NDJSON line from a truncated cache write) are dropped with a `-003` diagnostic instead of crashing the loader and aborting sibling upstreams.
- Manifest-controlled data-file paths are containment-checked (absolute paths and `..` segments rejected) — closing the latent traversal analogous to #699 before slice 2 populates caches from remote sources.

## Tests

26 focused tests across `core/compiler/deserialize_test.ts` (16) and `core/upstream/mod_test.ts` (10), all through the real parse pipeline; full gate green (`just check`: 3612 tests at final count, deno check/lint/dprint clean).

## Working-memory note (accepted debt, visible per the wip-gate rule)

`docs/wip/` on this branch carries the epic's design spec + slice-1 plan. They remain live working memory for slices 2-6 of this epic; gardening into durable records (ADR + guide) is slice 6's explicit deliverable. Reason: archiving the spec now would orphan the 5 remaining slices that execute against it.

### DIFF
--- a/docs/wip/2026-07-04-federated-upstream-resolution-design.md
+++ b/docs/wip/2026-07-04-federated-upstream-resolution-design.md
@@ -1,0 +1,518 @@
+# Federated upstream resolution — design
+
+- **Date:** 2026-07-04 (revised same day after org-SSOT alignment)
+- **Status:** draft, awaiting spec review
+- **Scope:** true peer-to-peer cross-repository references — a downstream repo's
+  entry traces (`Satisfies:`, `Derived-from:`, …) to an entry authored in
+  another repository, resolved and validated by `markspec check`; plus the
+  program root project that aggregates the overall picture.
+
+## 1. Problem
+
+MarkSpec today has exactly one working cross-repo entry-sharing mechanism:
+profile-delivered corpus (ADR-030), which ships a shared baseline _inside a
+profile package_. Peer-to-peer references — repo B's requirement satisfying a
+live entry in repo A — do not resolve: the federation protocol is designed
+(compile-output spec §5, `language.md` §3.2, issue #22) but the resolution side
+was never built. `resolveRegistries` pins an upstream's manifest hash into
+`markspec.lock`, yet nothing ever reads an upstream's entries.
+
+Target topology is a mix: product→component references up the V-model, peer
+repos (ICDs), shared standards resolution (RefHub-style registries), and a
+root/program repo that needs the overall picture.
+
+## 2. Prior art this design builds on
+
+| Piece                             | Where                                          | Reused as                                         |
+| --------------------------------- | ---------------------------------------------- | ------------------------------------------------- |
+| Org project-manifest contract     | `driftsys/schemas` `project/v1.json` + README  | The declaration surface (SSOT for `project.yaml`) |
+| Lockfile + `resolveUpstreams`     | ADR-022, `core/lock/`                          | Pinning + drift machinery; new upstream row kind  |
+| Registry pinning                  | `core/lock/resolve.ts::resolveRegistries`      | Extended from pin-only to pin+snapshot            |
+| `Entry.origin` read-only citizens | ADR-030, `core/profile/delivered.ts`           | Same model, new origin kind `upstream`            |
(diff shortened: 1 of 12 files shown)


stasson:
## /review findings (8-angle sweep, 1-vote verified)

Correctness angles (line-by-line, removed-behavior, cross-file trace) returned **no findings** — consistent with the per-task review loops and whole-branch review this branch already went through. Four cleanup-grade findings survived verification; none block merge. Recommended disposition in parentheses.

1. **`core/upstream/mod.ts:135-147` — path-containment predicate duplicated, and enforced by convention only.** `isUnsafeRelPath` reimplements (regex form) the exact predicate `PROFILE-DELIVERS-003` already implements in `core/profile/manifest.ts:1395-1401` (normalize+split form) — two copies of a *security* predicate that must not drift. And the guard lives at the one current call site: `extractSerializedEntries` is exported with no path-safety in its contract, so slice 2's `references:` fetcher (next manifest-controlled-path consumer) must remember to re-apply it. Fix: extract one shared containment helper (e.g. `core/util/`) used by both sites, and/or enforce inside `extractSerializedEntries` where the path is consumed. (**Fix in slice 2 when the second caller appears — flagged as an acceptance criterion there.**)
2. **`core/model/mod.ts:587-591` — stale `Entry.origin` doc comment.** Still reads "Set on entries injected from a profile-delivered corpus document (ADR-030)"; this PR widens the union to `upstream` origins hydrated by `loadUpstreamCorpus`. A slice-2/4 implementer reading it could write `kind === "profile"`-specific read-only gating and silently exempt upstream entries. (**One-line fix, worth taking in this PR.**)
3. **`core/compiler/deserialize.ts:12` ↔ `core/mod.ts` — two-file import cycle vs AGENTS.md "no cycles" rule.** `deserialize.ts` imports `CORE_SCHEMA_VERSION` from the barrel, which re-exports `deserialize.ts`. Verified benign (function/const bindings, existing precedent in `discipline_classifier.ts` — the constant is *declared in* the barrel, so "import from the defining module" was unsatisfiable). Known fix: relocate `CORE_SCHEMA_VERSION` to a leaf module (`core/model/` or `core/version.ts`). (**Defer to slice 2, which touches manifest/lock plumbing.**)
4. **`core/compiler/deserialize_test.ts` + `core/upstream/mod_test.ts` — 9-field manifest fixture duplicated verbatim.** The two literals can drift on a future manifest-shape change. (**Defer / opportunistic — a shared test-fixture builder when either file is next touched.**)

Also deferred with reasons recorded in the branch ledger: diagnostic `location.file` pointing at `manifest.json` for data-file errors (revisit when diagnostics surface to users in slice 4); NDJSON multi-pass extraction (cold path, bounded impact).

Issue #726, opened 2026-07-03
story(uxil): compiler + uxRegistry + machine projection


stasson:
Part of #717 §B. Depends on: S7 (uxil parser). The heart of uxil — semantics + corpus registry.

## Goal
Compile parsed uxil into the corpus **`uxRegistry`** (parallel to `typeRegistry`) + a machine-readable projection (downstream codegen/validators entry point).

**Closed, core-owned vocabularies** (extension = a markspec release decision):
- surface **kinds**: `screen`, `panel`, `agent`
- interaction **verbs**: `activate`, `toggle`, `select`, `adjust`, `input`, `scroll`, `drag`, `navigate`, `dismiss`, `ask`, `observe`

**Enforced semantics:**
- hierarchy inferred from dot-nesting; every surface declared exactly once corpus-wide;
- `navigate ->` targets resolve to a declared surface of a navigable kind;
- state bindings reference declared states; `!verb` citations check membership in the element's declared verb set;
- kind-specific rules (exclusivity, payload shape, navigability, statefulness).

## Acceptance
- `uxRegistry` built from a multi-entry corpus; machine projection deterministic + documented.
- All enforced semantics covered (declared-once, navigate resolution, state/verb membership, kind rules) — diagnostics wired via S9.

stasson:
Handoff note from #780 / PR #796 (S7 parser diagnostics cleanup), for the S8 implementation:

`parseRootDecl` now returns a **best-effort partial decl** on a missing kind — `{ kind: "", surface, states }` alongside a `UXIL-004` diagnostic — instead of `decl: undefined` (rationale: mid-edit editor tooling keeps the surface while the squiggle points at the kind). `RootDecl.kind`'s type cannot distinguish `""` from a real kind, so:

**S8's assemble must gate uxRegistry registration on a clean parse (empty `diagnostics`), never on `decl` truthiness alone.** The natural `if (decl) register(...)` pattern would register kind-less surfaces.

Same applies in spirit to `parseElementBullet` (returns a decl with `verbs: []` next to UXIL-005). The contract is also recorded in `parseRootDecl`'s doc comment; the S7 design doc §9 predates it.

PR #749, opened 2026-07-04
feat(core): typl published/namespaced tier (#723)

### LINKED ISSUE #723: story(typl): published/namespaced tier (dotted paths, declared-once, retire TYPL-002/003)
Part of #717 §C.1. Depends on: S4 (base-resolution). ⚠️ **Backward-compat gate — see below.**

## Goal
Add a **published** typl tier beside the unchanged entry-local tier:
`$powertrain.brake.pedal_position : signal float[0..100]` — dotted path, declared **exactly once corpus-wide**, citable from any entry (absolute, or relative under a base per §A). Retires the flat-global pairwise-consistency model (**TYPL-002/003**) for single-declaration ownership + a duplicate-declaration diagnostic.

## ⚠️ Compatibility gate (mandatory)
Dots are not part of a `$name` today (`IDENT_BODY_RE=/[A-Za-z0-9_]/` in typl/lexer.ts; `ENTITY_REF_RE=/\$([A-Za-z][A-Za-z0-9_]*)/` in parser/body_tokens.ts). To keep existing content working:
- **Recognize dotted paths only inside code-span / declaration context** (§A rule 6) so free-prose `$foo.bar` keeps the old tokenization (no re-tokenization, no new Q500 false positives).
- **Deprecate — do not hard-delete — TYPL-002/003** during transition (removing errors only relaxes; keep codes resolvable for tooling).
- Bump **CORE_SCHEMA_VERSION**: compiled output gains the published tier and `typeRegistry` semantics change (downstream RIDL emitters must adapt).

## Acceptance
- Published dotted symbols: declared-once diagnostic on duplicate; cite from other entries (absolute + relative-under-base).
- Entry-local plain `$Name` behavior unchanged (regression proof).
- Free-prose `$foo.bar` unchanged (compat regression test).
- Schema bump + release note (no migration tooling per policy).

stasson:
Closes #723.

## Summary

Adds a **published** typl tier beside the unchanged entry-local tier: dotted
paths (`$powertrain.brake.pedal_position`, ≥2 segments) are declared exactly
once corpus-wide and citable from any entry, absolute or relative-under-base
(`$.pedal_position`). An explicit `` `$name : namespace` `` clause creates a
base; relative refs resolve through the shared entry-local base-resolution
engine (`core/decl/resolve.ts`, #722). Citations (bare published-shaped code
spans) are validated against the corpus registry.

- New diagnostics: **TYPL-009** (duplicate published declaration), **TYPL-010**
  (relative ref with no base in scope), **TYPL-011** (citation of an undeclared
  symbol), **TYPL-012** (multiple root namespaces in one entry).
- **TYPL-002/003 retired** (deprecated, never emitted — kept resolvable for
  tooling): plain (entry-local) `$Name` no longer has a cross-entry
  pairwise-consistency rule, since publication now requires explicit dotted
  namespacing.
- Free-prose `$foo.bar` tokenization (`core/parser/body_tokens.ts`) is
  untouched — dotted paths are recognized only inside declaration/citation
  context (code spans / bullet declarations), never in plain prose.
- Durable record: `docs/architecture/adr-019-typl-type-dsl.md` addendum
  ("published tier, #723").

**Deliberate deviation from the issue's literal compat gate:** the issue asked
for a `CORE_SCHEMA_VERSION` bump. Investigation during design (recorded in the
now-archived `docs/archive/specs/2026-07-04-typl-published-tier-design.md`,
Decision 1) found the bump isn't load-bearing — nothing gates on the constant,
`typeRegistry` already serializes as `{}` via `JSON.stringify` on the in-memory
`Map`, and no RIDL emitter exists yet to depend on wire semantics — so it's
dropped. Caveat recorded for when a future RIDL/payload-bridge emitter needs
real serialization.

## Release note (batch at release time)

> Added a published typl tier: dotted names (`$a.b.c`) declare a corpus-wide
> symbol exactly once and are citable from any entry, with an explicit
> `: namespace` base clause and relative (`$.x`) refs. TYPL-002/003 (cross-entry
> plain-name consistency) are retired — entry-local `$Name` no longer has a
> cross-entry rule. No schema-version bump; no migration required.

## Test plan

- [x] 3616 unit/integration tests pass (`just build`)
- [x] 9 new e2e acceptance tests (`tests/e2e/typl_published_test.ts`) covering
      declare-once, citation (absolute/relative), TYPL-009/010/011/012, and
      compat regressions (entry-local mismatch silent, free-prose untouched)
- [x] Sibling e2e suites (`typl_validation_test.ts`, `typl_bullet_test.ts`,
      `typl_fence_test.ts`, `typl_inline_test.ts`) updated for the TYPL-002/003
      retirement and passing
- [x] `deno lint`, `deno fmt --check`, `dprint check`, `deno check` all clean
- [x] `docs/wip/` gardened to `docs/archive/{plans,specs}/`


stasson:
## Automated review (medium effort — 8 finder angles, 1-vote verify)

This PR adds a published typl tier (dotted `$a.b.c` names, declared once corpus-wide, citable via absolute or relative-under-base refs) beside the entry-local tier, retiring TYPL-002/003 in favor of TYPL-009/010/011/012. 8 candidate findings were verified by independent agents (each re-read the code and, where possible, ran a reproduction); one (root-only citation scope) was **REFUTED** — it's the explicitly planned v1 design, not a bug. The 8 that survived, most severe first:

1. **`packages/markspec/core/typl/validator.ts:48`** — Retiring TYPL-002/003 also silently drops the **same-entry** consistency check, not just the cross-entry one. Before this PR, `registry.bindings` grouped by `$Name` corpus-wide with no entry distinction, so two bullets *in the same entry* declaring `$speed` with conflicting kinds (`signal` vs `state`) already fired TYPL-002 incidentally. The new loop's `if (!name.includes("."))  continue;` drops all plain names — including that same-entry case — before any check runs. Verified by running the identical repro on `main` (exits 1, `TYPL-002`) vs this branch (exits 0, no diagnostics). This contradicts the PR's own stated acceptance criterion, "entry-local plain `$Name` behavior unchanged (regression-proof)."

2. **`packages/markspec/core/typl/assemble.ts:65`** — Multi-level namespace nesting silently breaks for the standard backtick-wrapped bullet convention used in every doc example and the acceptance tests. Root cause: bullet paragraph text is stored verbatim (backticks included), so `isTyplDeclarationText`'s `^\$` anchor rejects backtick-wrapped bullets from the nesting-aware bullet surface; they fall through to the flat inline-code surface instead, which carries no parent link. Verified by reproduction: a 3-level nested, fully backtick-wrapped example (`$powertrain:namespace` → `$.brake:namespace` → `$.pedal:signal`) resolves to `$powertrain.pedal` (silently skipping `brake`) with zero diagnostics; spelling the inner namespace absolutely instead produces a false TYPL-012. An identical bare-bullet (no backticks) version resolves correctly.

3. **`packages/markspec/core/typl/assemble_test.ts` (~L31)** — Directly explains why #2 wasn't caught: the only unit test for "innermost wins" nesting deliberately uses bare, non-backtick syntax, with a comment admitting backtick-wrapped bullets route through a different, nesting-blind surface. No test anywhere (unit or e2e) exercises a 2+-level nested namespace in the backtick form every real example uses.

4. **`packages/markspec/core/typl/resolve.ts:15`** — `isRelativeTyplName` accepts a digit-leading segment (`$.1abc`) but `recognize.ts`'s `BINDING_RE` (which gates whether text is recognized as a declaration at all) rejects it. Since `citations.ts` uses exactly these two predicates with no stricter filter, a citation shaped `` `$.1abc` `` is accepted as citable but can never be declared — it's permanently unresolvable (TYPL-010/011), a dead end baked into the grammar.

5. **`packages/markspec/core/typl/assemble.ts:120`** — TYPL-012's "at most one root" rule is hand-rolled instead of calling `core/decl/resolve.ts`'s `checkSingleRoot`, which was built for exactly this rule (doc comment names it explicitly) and has zero callers anywhere outside its own test. The PR's own design doc (§5) states in the present tense that `checkSingleRoot` "is applied" here — that claim doesn't match the shipped code.

6. **`packages/markspec/core/typl/validator.ts:44`** — The TYPL-009 "declared exactly once corpus-wide" rule is inlined rather than generalized into `core/decl/`. Not speculative: issue #726 (S8, same open epic, next phase after this PR) explicitly requires the identical "declared exactly once corpus-wide" semantic for uxil's registry — this logic will need to be duplicated there imminently.

7. **`packages/markspec/core/compiler/mod.ts:362`** — `compile()` was already calling `validateTypl` twice (pre-existing structure, unrelated to this PR), discarding the second call's diagnostics just to get `.registry`. This PR adds a citation-validation loop (walks every entry's `bodyTokens`, `resolveRef` + registry lookup per citation) directly into `validateTypl`, so that new, heavier work now runs twice per `compile`/`show`/`context`/`report`/`export` call. `buildTypeRegistry` (already used standalone in `lsp/workspace.ts`) would produce the same registry without re-running validation.

8. **`packages/markspec/core/typl/bullet.ts:31`** — `extractTyplBullets` (and the `core/decl/extractBulletDeclarations` it wraps) has zero production callers left after this PR's own refactor moved the parser onto `extractTyplBulletsNested`; only its own test still exercises it. Dead code.

*(Dropped from the cap: a missing doc comment on `assembleTyplTypes` — real but the lowest-severity item once 8 correctness/altitude/efficiency findings were confirmed.)*

PR #756, opened 2026-07-04
feat(lsp): align typl hover and completion with the published tier

### LINKED ISSUE #750: typl LSP alignment: hover/completion for the published tier (#723 follow-up)
Part of #717. Follow-up from the #723 design review (S5, "typl published/namespaced tier").

## Problem

#723 (PR #749) adds a **published** typl tier: dotted names (`$powertrain.brake.pedal_position`) declared exactly once corpus-wide, plus relative refs (`$.x`) resolved through an entry-local base-resolution engine. The LSP layer (`packages/markspec/lsp/typl.ts`, wired into `lsp/server.ts`'s hover and completion handlers) was intentionally left untouched by #723 (deferred — see the design record, now archived at `docs/archive/specs/2026-07-04-typl-published-tier-design.md`) and still assumes the old flat-global registry model:

- `dollarNameAtPosition` (`lsp/typl.ts`) tokenizes a `$Name` under the cursor with a pattern that does not span dots — hovering or completing on a dotted published name (`$powertrain.brake.pedal_position`) won't recognize the full token.
- Hover content built from the registry (`formatTyplHoverContent`) doesn't know about the declared-once / citation-vs-declaration distinction, so hovering a plain `$Name` that happens to collide across entries (now legal — TYPL-002/003 are retired) shows misleading "cross-entry" framing.
- Completion (`buildDollarNameCompletions`) lists registry names flat, without namespace-aware grouping or relative-ref (`$.`) suggestions scoped to the entry's root namespace.

## Proposed scope

- Extend `dollarNameAtPosition`'s token grammar to span dots (and the relative `$.` form).
- Teach hover to distinguish a declaration site from a citation, and to render a published symbol's full dotted path + declaring file.
- Scope completion suggestions: inside an entry with a root namespace, offer relative (`$.x`) completions in addition to absolute dotted names.

## References

- #723 / PR #749 (published tier implementation)
- Design record: `docs/archive/specs/2026-07-04-typl-published-tier-design.md` (design review note on deferred LSP alignment)

stasson:
Closes #750.

Follow-up to #723 / PR #749 (published typl tier). The LSP typl layer
(`lsp/typl.ts`, wired into `lsp/server.ts` hover + completion) still assumed the
pre-#749 flat-global registry; this aligns it with the published tier. **No core
change** — the registry already keys published leaves by their fully-qualified
absolute name and carries the entry `rootNamespace`.

## What changed

**Tokenizer — `dollarNameAtPosition`**

- Token grammar widened to span dots, so dotted published names
  (`$powertrain.brake.pedal_position`) and the relative `$.x` form are single
  tokens.
- The `$` sigil breaks the left scan (`path.to.$sig` → `$sig`); a cursor on a
  bare `.` separator returns nothing; a trailing sentence period is trimmed
  (`$a.b.` → `$a.b`); bare `$` / `$.` → nothing.

**Hover — `formatTyplHoverContent(name, registry, context?)`**

- New optional `context` carrying the enclosing entry's display ID + root
  namespace.
- Relative `$.x` resolves against the entry **root** namespace only.
- Published `$a.b` renders full dotted path, kind, shape, and its declaring
  entry/file — or "declared in this entry" when the cursor is inside the
  declaring entry.
- Plain `$Name` drops the retired TYPL-002/003 "collision" framing; two entries
  declaring the same plain name are shown as independent entry-local symbols.

**Completion — `buildDollarNameCompletions` + triggers**

- `$.` (relative) offers `$.tail` shorthands for leaves under the entry root
  namespace; plain `$` offers the flat corpus list **plus** this entry's
  `$.tail` shorthands.
- `isDollarNameTrigger` widened to fire on `$`, `$.`, `$.ped`, `$a.b`; new
  `isRelativeDollarTrigger`.

## Design decisions

- **Root-only relative resolution** (agreed with reporter). Nested-namespace
  bases are not persisted per line, so the LSP resolves a relative `$.x` against
  the entry root only. A `$.x` under a *nested* namespace shows **no** hover
  popup rather than a wrong one — never an incorrect answer. Matches the issue's
  "scoped to the entry's root namespace" wording. Nested-scope reconstruction is
  out of scope.
- Label-only completion insertion kept unchanged (no `textEdit` plumbing), so
  absolute-insert behavior is untouched.

## Tests

27 colocated unit tests in `lsp/typl_test.ts` (tokenizer dotted / relative /
trailing-dot / dot-separator / prose cases; hover published / relative /
entry-local-no-collision / "this entry"; completion relative + absolute
scoping; triggers). Existing tests stay green. Full `just check`: **3669
passed, 0 failed**.

## Noted (out of scope)

`docs/guide/typl.md` still documents TYPL-002/003 as active diagnostics and has
no published-tier section — pre-existing doc debt from #723/#749. This PR only
updated the LSP "Editor support" bullets its own behavior changes; a broader
published-tier guide pass is worth a follow-up.

### DIFF
--- a/docs/guide/typl.md
+++ b/docs/guide/typl.md
@@ -216,10 +216,13 @@ need a flat lookup table.
 
 The MarkSpec LSP provides two typl affordances when you open a Markdown file:
 
-- **Hover** — hovering over a `$Name` token shows its kind, shape, and which
-  entry declared it.
-- **Completion** — inside a `typl` fence or after `$`, the LSP offers names
-  already declared in the workspace as completion candidates.
+- **Hover** — hovering over a `$Name` token shows its kind, shape, and where it
+  is declared. A dotted published name (`$powertrain.brake.pedal_position`)
+  shows its full path and declaring file; a relative `$.name` reference resolves
+  against the entry's root namespace.
+- **Completion** — inside a `typl` fence or after `$`, the LSP offers the
+  workspace's declared names. Inside an entry that declares a root namespace,
+  typing `$.` offers relative shorthands for that namespace's symbols.
 
 Both features require the `markspec lsp` server to be running. In VS Code,
 install the MarkSpec extension — it starts the server automatically.
--- a/packages/markspec/lsp/server.ts
+++ b/packages/markspec/lsp/server.ts
@@ -139,6 +139,7 @@ import {
   dollarNameAtPosition,
   formatTyplHoverContent,
   isDollarNameTrigger,
+  isRelativeDollarTrigger,
 } from "./typl.ts";
 import { buildVersionNotification } from "./version_notification.ts";
 
@@ -1041,7 +1042,17 @@ connection.onCompletion((params): CompletionItem[] | CompletionList => {
   if (isDollarNameTrigger(line)) {
     return time("onCompletion/dollarName", () => {
       const registry = index.getTypeRegistry();
-      const items = buildDollarNameCompletions(registry);
+      // Published tier (#750): scope relative `$.x` shorthands to the
+      // enclosing entry's root namespace, and switch to relative-only
+      // suggestions when the partial is `$.`-led.
+      const enclosing = findEnclosingEntry(
+        index.getEntriesForFile(filePath),
+        params.position.line + 1, // LSP is 0-based; Entry is 1-based.
+      );
+      const items = buildDollarNameCompletions(registry, {
+        rootNamespace: enclosing?.types?.rootNamespace,
+        relative: isRelativeDollarTrigger(line),
+      });
       return items.map((item) => ({
         label: item.label,
         detail: item.detail,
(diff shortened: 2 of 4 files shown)


stasson:
## Post-PR review

Reviewed the diff for correctness, regression, cleanup, and conventions (medium effort). **No findings survived verification.**

Key candidates checked and refuted:

- **Tokenizer left-scan swallowing prose** (`typl.ts`) — refuted: `$` is excluded from `TOKEN_CHAR_RE`, so the left scan stops at the sigil; a dotted prose chain with no `$` hits `start === 0 || line[start-1] !== "$"` → `undefined`. Traced `path.to.$sig → $sig`, `path.to.thing → undefined`, `$.pedal` (cursor mid-token), bare `$`, `$.`, and trailing-dot cases — all correct.
- **`findEnclosingEntry` off-by-one** (`server.ts` hover + completion) — refuted: both new sites pass `params.position.line + 1`, identical to the pre-existing trace-attribute completion, and `find_entry.ts` documents its param as 1-based matching `Entry.location.line`.
- **`isDollarNameTrigger` widening stealing a trigger** — refuted: the dollar trigger is the last trigger in `onCompletion`, so widening it can't pre-empt earlier ones; new firings are dotted partials, and the client filters the returned list (harmless empty popup on e.g. `$5.00`).
- **Caller regression on new optional params** — refuted: `formatTyplHoverContent` / `buildDollarNameCompletions` are called only from `server.ts`; both added params are optional with defaults.
- **Relative-hover under a nested namespace** — matches the deliberate root-only design (graceful miss, never a wrong answer), not a bug.

Non-actionable note: absolute-mode completion makes two registry passes, but the second early-returns when the entry has no root namespace — negligible on a user-triggered, client-filtered completion.

Gate: full `just check` green (3669 passed, 0 failed); 27 colocated typl unit tests.

stasson:
Filed the noted published-tier doc debt as #758 (retire TYPL-002/003 in the guide, add a namespaced-declarations section).

Issue #741, opened 2026-07-04
Epic: Federated upstream resolution — cross-repo trace references


stasson:
**Epic** · label `epic:federated-upstream`

## Summary

True peer-to-peer cross-repository trace references: a downstream repo's
entry traces (`Satisfies:`, `Derived-from:`, …) to an entry authored in
another repository, resolved and validated offline by `markspec check`
against a lock-mediated snapshot cache — plus a root/program repo pattern
that aggregates the overall multi-repo picture.

Design: `docs/wip/2026-07-04-federated-upstream-resolution-design.md`
(spec + slice-1 plan committed to `main` as accepted debt in #740's body;
gardening into `docs/decisions/` + the guide is slice 6's deliverable).

## Context

MarkSpec has one working cross-repo entry-sharing mechanism today:
profile-delivered corpus (ADR-030), which ships a shared baseline *inside a
profile package*. Peer-to-peer references — repo B's requirement satisfying
a live entry in repo A — do not resolve: the federation protocol was
designed (compile-output spec §5, `language.md` §3.2, #22) but the
resolution side was never built.

Key decisions (D1–D10 in the design doc):

- Declaration surface = org SSOT `driftsys/schemas` `project/v1.json`:
  `dependencies:` (git repos, full traceability + coverage) and
  `references:` (published sites, leaves). `parents:` / `parent-fallback:`
  are retired; RefHub becomes an explicit `references:` entry.
- **Lock-mediated network**: only `markspec lock` touches the network;
  `check` / `compile` / LSP / MCP resolve offline from the pinned cache.
  `projectRef.version` carries intent (absent = auto: latest tag else
  branch head; tag = frozen; branch = track) — the lockfile records the
  resolved kind (`tag` \| `branch` \| `sha`).
- Upstream entries join the graph as read-only citizens via a new
  `EntryOrigin` kind (`upstream`), reusing the ADR-030 model. Flat display-ID
  space; any collision is a hard `MSL-R014`-shaped diagnostic.
- **Authoritative-source rule**: hydration skips snapshot entries that
  already carry an `origin`, so a root aggregating diamond-shaped
  dependencies never double-counts — this is what makes the root/program
  pattern (§4.9) collision-free.
- Release assurance is two-level: pin-level (`check --strict` requires every
  dependency tag-resolved) and content-level (a moved pin can make a
  previously-resolving reference fail once the tag doesn't contain it).
- Core stays pure (D10) — fetchers (git, HTTP, tarball) live at the
[...]

PR #757, opened 2026-07-04
feat(core)!: federated upstream slice 2 — org manifest adoption + lock-mediated references

### LINKED ISSUE #742: story(core): org manifest adoption + lock (federated upstream slice 2)
Part of #741 (slice 2 of 6). Depends on slice 1 (shipped, `60ceec5`).
Design: `docs/wip/2026-07-04-federated-upstream-resolution-design.md` §4.1,
§4.2, §4.10.

**External prerequisite:** `driftsys/schemas#6` (document the
`projectRef.version` intent contract) should merge first — this story reads
that contract literally.

## Goal

Adopt the org-schema `project.yaml` contract (`driftsys/schemas`
`project/v1.json`) as the declaration surface for cross-repo relationships,
and extend `markspec lock` to resolve, snapshot, and pin them offline
thereafter.

## Scope

- Org-schema `project.yaml` loader: `dependencies:` and `references:` lists
  of `projectRef` (`url` required; `name`, `version` optional). Retire the
  local `schemas/project/v1.json` — the org schema is SSOT (`name` +
  `version` become required).
- Migrate markspec-only tool config off `project.yaml` onto
  `.markspec.yaml`: `exclude:`, `caption-conventions:`, and `labels:`
  vocabulary (all disallowed by the closed org schema — see design §4.10's
  table).
- Reference fetcher (`https://` and `file://` URLs) for `references:`
  entries — fetch the published compile-output manifest, write the
  content-addressed snapshot cache under
  `.markspec/cache/upstreams/<id>/`.
- Lockfile rows extending the ADR-022 `[[upstream.*]]` family
  (`[[upstream.dependency]]` / `[[upstream.reference]]` — exact TOML naming
  is plan-time detail per design §10) recording `intent`, `resolved`, `sha`
  (dependencies only), `snapshot` hash, `locked-at`.
- Three lock flows, all offline-safe outside `markspec lock` itself:
  first lock, restore (pins exist, cache doesn't — fetch the pinned
  sha/snapshot, verify hash, error on mismatch), update
  (`lock --update[=<id>]` — re-resolve intent, move pin if changed).
- `MSL-L212` drift gate gains one case: a locked upstream whose cache
  snapshot is missing or hash-mismatched.
- Reconcile the org repo's `markspec/lock/v1.json` ("frozen sidecar
  `.markspec.lock`") naming/shape against this repo's ADR-022
  `markspec.lock` before extending lockfile rows (design §9 #3).

Git-repository acquisition itself (the tarball/clone fetcher for
`dependencies:`) is slice 3 — this slice's lock flows apply to
`references:` end-to-end and lay the manifest/lockfile groundwork
`dependencies:` reuses.

## Out of scope

- Actual git-repo cloning/tarball fetch for `dependencies:` (slice 3).
- Graph hydration, validator wiring, `MSL-T014` (slice 4).
- `kind` discriminator on `projectRef` (deferred — design §5 #8, §8).

## Testing

- Unit: org-manifest loader against valid/invalid `projectRef` shapes.
- E2E: two temp projects — compile A to `/api/`, point B's `references:`
  at it via `file://`; assert `lock` pins, snapshot cache is written, and a
  deleted/corrupted cache trips the `MSL-L212` case.

stasson:
Closes #742.

Slice 2 of the federated-upstream epic (#741) — org manifest adoption + the
lock-mediated references flow. Design:
`docs/wip/2026-07-04-federated-upstream-resolution-design.md` (§4.1 declaration
surface, §4.2 lock flows/rows, §4.4 release assurance, §4.10 schema SSOT +
tool-config migration). Executed against the slice-2 plan
(`docs/wip/2026-07-04-federated-upstream-slice2-plan.md`); each of the 11 tasks
was implemented TDD-first and reviewed, then a whole-branch review ran.

## What ships

**Org project.yaml contract (closed schema).** `project.yaml` now validates
against the org contract (`https://driftsys.github.io/schemas/project/v1.json`):
`name` (reverse-domain `^[a-z][a-z0-9.-]*$`) and `version` are required, and
new `dependencies:` / `references:` are lists of projectRef `{ url, version?,
name? }`. Unknown keys are rejected.

**Tool-config migration (breaking, pre-1.0 — no back-compat).** `exclude:` and
`caption-conventions:` move from `project.yaml` to `.markspec.yaml`;
`labels:`-as-vocabulary, `parents:`, and `parent-fallback:` are retired (RefHub
becomes an explicit `references:` entry). Removed keys produce a clear
migration error, not silent acceptance. The local `schemas/project/v1.json` is
retired for the org SSOT; the two migrated keys move into
`schemas/markspec/v1.json`.

**Lock-mediated references.** `markspec lock` resolves `references:` over https
and `file://`, caching the published compile-output snapshot under
`.markspec/cache/upstreams/<id>/` (gitignored) and pinning it in an extended
`[[upstream.registry]]` row with a `snapshot` hash. Three flows: first-lock,
restore (repopulate a broken cache, verified against the pin — mismatch is
`MSL-L214`), and update (`--update` / `--update=<id>` moves the pin). A pinned
reference resolves fully offline; only `lock --update` (or a broken cache)
touches the network. `markspec check` gains an offline cache-drift case on
`MSL-L212` (missing/mismatched cache → run `markspec lock`); a reference that
can't be locked is `MSL-L213`. The compile manifest now records
`project.version` and sources `federation` from `references:`.

**Feeds slice 1.** The lock-written cache is exactly the input slice 1's
`loadUpstreamCorpus` reads — proven by an integration test and an end-to-end
`file://` scenario suite (lock → cache-drift → restore → restore-mismatch →
update, plus the migration errors).

## Scope boundaries

- Git `dependencies:` are **declared but not yet acquired** — the row kind,
  parser/serializer, and a declared-not-written notice land here; the
  ls-remote + tarball fetcher is slice 3. The guide states this so authors
  don't expect resolution.
- Graph feed sites, `MSL-T014`, the generalized `MSL-R014` collision pass, and
  coverage-by-field are slice 4; LSP/MCP surfaces slice 5; the ADR + full
  `language.md`/compile-output-spec rewrites are slice 6. Only the user guide
  is updated here.
- A `references:` `version:` is recorded but not yet used to select a
  published baseline (documented as not-yet-enforced); it becomes load-bearing
  for git dependencies in slice 3.

## Carried slice-1 review criteria (all closed)

1. Shared `isUnsafeRelPath` path-containment helper (`core/util/paths.ts`) now
   backs both the upstream cache guard and PROFILE-DELIVERS-003 — and was
   hardened for Windows drive-relative/UNC forms.
2. `CORE_SCHEMA_VERSION` moved out of the `core/mod.ts` barrel to the model
   leaf, killing the `deserialize.ts` ↔ barrel import cycle.
3. `ReadFile` deduped — the `core/upstream` local copy now re-exports the
   canonical `core/config` type.
4. The `markspecSchemaVersion` literal deduped into `MANIFEST_SCHEMA_VERSION`.
5. Org lock-schema naming reconciliation was resolved as out-of-repo — the
   org `markspec/lock/v1.json` describes a different artifact (`.markspec.lock`
   per-entry frozen metadata); filed as driftsys/schemas#7. This branch
   extends the ADR-022 `markspec.lock`.

## External prerequisite (landed)

driftsys/schemas#6 documents the `projectRef.version` intent contract (exact
tag = frozen baseline, branch = track head, absent = auto) and the per-field
URL guidance — the contract this slice implements against.

## Verification

`just build` green: **3663 passed, 0 failed, 2 ignored**; `deno fmt --check`
and `dprint check` both clean. The whole-branch review returned
ready-to-merge with no Critical/Important findings.


stasson:
## `/review` — federated upstream slice 2

Automated review (8 finder angles → verify). The branch is well-built and type-checks cleanly across all four entry points; no convention violations and no hard call-site breaks. The migration re-establishes every removed guard I traced. Findings below are ranked most-severe first — the first four are worth acting on before merge, the rest are notes.

### 1. Spurious `MSL-L214` aborts every lock when a matching registry row lacks a `snapshot` — `packages/markspec/core/lock/upstream_refs.ts:272`
`cacheIntact` returns `false` for a snapshot-less row (`upstream_refs.ts:206`), so the keep flow refetches and then evaluates `fetched.snapshotHash !== existing.snapshot`. When `existing.snapshot === undefined`, that comparison is always true → `MSL-L214` "the published site moved" (severity error) → the lock aborts (see #2) and never self-heals except via `--update`. The parser round-trips snapshot-less `[[upstream.registry]]` rows (there's a test), so this state is representable. Fix: in the keep branch treat `existing.snapshot === undefined` as "no pin to verify → re-pin as first-lock", not a mismatch.

### 2. One unreachable `references:` entry blocks the entire `markspec lock` write — `packages/markspec/cli/commands/lock.ts:~202`
A failed manifest/data fetch is `MSL-L213` (severity error), and lock exits 1 **before writing `markspec.lock` at all** — so a transient 404 or offline run also drops the project's own edges, profile chain, and bound entries, not just the failing reference. The removed `resolveRegistries` treated an unreachable manifest as an `MSL-L101` *warning* and still wrote the lockfile. This is the documented "never record a partial pin" choice, but the blast radius covers all non-reference content and can wedge CI (which needs the lockfile to exist for `--frozen`). Worth confirming the trade-off is intended vs. warn-and-write-the-rest.

### 3. `markspec check` does not detect a declared-but-unlocked `references:` entry — `packages/markspec/cli/commands/check.ts:192`
The composite `check` gate (the ADR-027 primary CI entry point) calls only `lockfileDriftGate` (edge-drift + offline cache-drift over *locked* rows). It never calls `checkDrift` / derives `declaredReferenceIds`, so `MSL-L202` ("declared reference has no lockfile row") fires only under `lock --check` and `compile --frozen`. A developer who adds a reference and forgets to re-lock passes `markspec check` clean. Blast radius is limited in slice 2 (upstream entries aren't fed to the graph yet) but grows in slice 4 — consider running the id-presence drift in `check` too, or documenting that `lock --check` is the required CI gate for references.

### 4. Cache-root path duplicated verbatim between the writer and the prober — `packages/markspec/cli/commands/lock.ts:189` & `packages/markspec/cli/commands/check.ts:175`
Both hardcode `join(projectRoot, ".markspec", "cache", "upstreams")` — `lock` writes it, `check` probes it, and they agree only by copy-paste. Change one and `check` probes a directory `lock` never populates → spurious project-wide `MSL-L212` on every locked reference. Extract one shared constant/helper in `core/lock`.

### 5. Pre-slice-2 lockfiles trip `MSL-L203` until a full re-lock — `packages/markspec/core/lock/check.ts`
`declaredReferenceIds` yields short ids (`refhub`) while old `resolveRegistries` rows are keyed `urn:markspec:registry:<url>`; the L203 loop flags every locked registry row not in the declared set, so `lock --check` / `compile --frozen` error until re-lock. Expected migration churn under the pre-1.0 no-back-compat policy (no committed lockfile in this repo), but worth a one-line migration note for downstream consumers.

### 6. `.markspec.yaml` is read and fully parsed twice per CLI invocation — `packages/markspec/cli/commands/lock.ts` (and check/compile/fmt/doctor)
Each command calls `loadProfileForCommand` (which parses `.markspec.yaml` for `profiles`) and then the new `loadToolConfig` (which parses the same file again for `exclude`/`caption-conventions`) — an extra read + full parse, including the profile-specifier validation loop. Parse once and derive both the chain and the `ToolConfig`.

### 7. Snapshot integrity hash covers only `entries.file`, not the ndjson `entries.index` — `packages/markspec/core/lock/upstream_refs.ts` / `cache_check.ts`
`snapshot` = sha256 of the entries data file only; a corrupt/truncated cached ndjson index passes both `cacheIntact` and the `MSL-L212` gate. Latent in slice 2 (references publish inline `compiled.json`), but revisit when slice 4 consumes the ndjson/index path.

### 8. Dead `?? []` on a non-optional field — `packages/markspec/core/compiler/manifest.ts:~93`
`(config.references ?? []).map(...)` — `ProjectConfig.references` is non-optional and `parseProjectRefList` always returns at least `[]`. Simplify to `config.references.map(...)`.

---
Also minor: `resolveUpstreams`' `config` option is now unused inside `resolve.ts` (still passed by both callers — harmless dead field), and the `MSL-L212` code is now triple-purposed (online edge-hash, offline edge, offline cache drift) — same remediation, but a machine consumer keyed on `code` can't distinguish the causes.

🤖 `/review` via [Claude Code](https://claude.com/claude-code)

stasson:
### CI fixed + two `/review` findings addressed

**Windows CI failure (root cause + fix).** The `Test (windows-latest)` job failed on the new `federated_lock_test.ts` step 1 while macOS/Linux passed. Root cause was a real Windows bug in the `file://` fetcher: `defaultFetchUrl` stripped the scheme with `url.replace(/^file:\/\//, "")`, which turns a Windows URL `file:///C:/…` into `/C:/…` — a path `Deno.readFile` can't open (POSIX tolerates the leading slash, Windows doesn't). Fixed by resolving the URL through `fromFileUrl()` (the std cross-platform primitive), and building the URL in the test via `toFileUrl(join(dir, "api")).href` instead of string concatenation (`7efc22c`). All 17 checks now green.

**Also addressed two findings from the review above (`f137a2e`):**
- #1 (spurious `MSL-L214`): the keep branch now re-pins a snapshot-less existing registry row as a first-lock instead of comparing against `undefined`; added a regression test.
- #4 (duplicated cache-root literal): extracted `upstreamCacheRoot(projectRoot)` in `core/lock`, used by both `lock` and `check`.

The remaining review items (#2 whole-lock abort on an unreachable reference, #3 `check` not catching a declared-but-unlocked reference) are design decisions left for discussion rather than changed unilaterally; #5–#8 are deferrable notes.

PR #759, opened 2026-07-04
fix(repo): make install.sh's temp-dir cleanup trap survive under set -u

stasson:
## Summary

`install.sh`'s `main()` declares `tmpdir` with `local` and registers an `EXIT` trap that references it (`trap 'rm -rf "$tmpdir"' EXIT`). The trap only actually fires after `main()` returns and the script falls off the end — at which point `tmpdir` has gone out of its local scope. Under `set -eu`, referencing the now-unbound `$tmpdir` aborts the script with `tmpdir: unbound variable`.

In practice this means: **every successful `install.sh` run currently exits non-zero** — the binary is correctly downloaded, verified, and installed, but the script then crashes during its own cleanup trap on the way out.

Fix: drop `local` so `tmpdir` stays in scope through to the trap's firing point, and default-expand it in the trap (`${tmpdir:-}`) as a defensive fallback for any other exit path.

## Test plan

- [x] Reproduced the bug in isolation with a minimal `sh` script mirroring the pattern (old pattern: exits 1 with "unbound variable"; new pattern: exits 0) before writing the fix
- [x] `shellcheck -s sh install.sh` — the same check CI runs (`.github/workflows/ci.yaml`) — stays clean
- [x] `just check` — 3669 passed, 0 failed

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -71,9 +71,11 @@ main() {
   echo "Installing markspec $version ($target)" >&2
   echo "  to: $INSTALL_DIR" >&2
 
-  local tmpdir
+  # `tmpdir` is intentionally global (not `local`): the EXIT trap fires at
+  # global scope, so a function-local would be unbound there under `set -u`
+  # and the cleanup would both error and leak the temp dir.
   tmpdir="$(mktemp -d)"
-  trap 'rm -rf "$tmpdir"' EXIT
+  trap 'rm -rf "${tmpdir:-}"' EXIT
 
   curl -fsSL "$url" -o "$tmpdir/$tarball"
   curl -fsSL "$checksum_url" -o "$tmpdir/$tarball.sha256"


stasson:
## `/review` — medium effort, 5 finder angles + verify

**What this PR does:** fixes `install.sh`'s `EXIT` trap referencing a `local` `tmpdir` that's already out of scope by the time the trap actually fires (after `main()` returns), which currently makes every successful install exit non-zero under `set -eu`.

**Findings: none survive verification.** Every finder independently reproduced the pre-fix bug (confirmed real: exit 1/2 with "unbound variable" on both bash and dash) and confirmed the fix closes it cleanly, with no regression on the `${tmpdir:-}` defensive fallback, no cross-file/CI dependency on `install.sh`'s internals, and no `shellcheck -s sh` issues.

**One non-blocking alternative worth knowing about, not requiring a change:** keeping `tmpdir` `local` (consistent with its `main()` siblings `target version tarball url checksum_url`) and instead using `trap "rm -rf '$tmpdir'" EXIT` (double-quoted, so `$tmpdir` is interpolated at *registration* time, not fire time) would sidestep the scope hazard entirely without introducing the local/global inconsistency the current fix has. Verified this alternative also works (exit 0, dir removed). Not requesting a change — the shipped fix is correct and the tradeoff is minor — just flagging it in case a future edit wants to revisit.

**Out of scope, filed separately:** the identical EXIT-trap/local-variable bug exists unfixed in `bootstrap`'s `ensure_git_std()` (`tmp_dir` at line 69/81) — there the `${:-}` guard was added but `tmp_dir` was never promoted out of `local`, so it doesn't crash, it just silently leaks the staging directory on every install/upgrade. Filed as #763.

PR #760, opened 2026-07-04
docs(docs): fix stale and fictional facts across the published guide + model book

stasson:
## Summary

Docs-accuracy sweep — revives and re-verifies an older, unlanded audit (its original branch had drifted 170+ commits behind `main`, so every fact was re-checked against current source rather than reapplied mechanically).

- **quickstart.md** Step 2: lead with `markspec init` (it ships — `ai-agents.md` already documents it) instead of a "coming soon" callout that contradicted it.
- **iso26262.md recipe + cli.md**: `@driftsys/iso26262` doesn't exist anywhere in the codebase (verified via grep across profile loading/tests) — reframed as a schema-valid local compliance profile example, matching `ADR-008`'s manifest structure.
- **iso26262.md recipe**: the "declare labels in `project.yaml` to enable MSL-L010 enforcement" claim is false — `MSL-L010` is an unrelated lockfile-reference code, and no code path enforces entry `Labels:` values against a profile-declared list. Replaced with the real benefit: `profile.labels:` drives known-label semantic highlighting in the LSP (`lsp/semantic_tokens.ts`, `lsp/entry_ranges.ts`).
- **editor-vscode.md** troubleshooting: "`markspec check` requires a `project.yaml`" is only true for bare invocation, not explicit-file invocation — replaced with the actual per-file recognition gate (`isMarkspecFile`/`isDocCommentContext`).
- **ci.md**: bumped the stale cache-key version placeholder (`0.6.0` → `0.10.3`) and noted it must move with the pinned/latest version to avoid serving a stale cached binary.
- **model/profiles.md**: replaced a dead link to an unpublished, gitignored internal design-spec file with the inline discipline-mode inference algorithm (`core/profile/discipline_mode.ts`'s `inferDisciplineMode`) it was pointing at.

## Test plan

- [x] Every fact independently verified against current source (`packages/markspec/core/profile/*`, `core/validator/listing.ts`, `lsp/semantic_tokens.ts`, `lsp/context.ts`, `deno.json` version) — not reapplied from the stale draft
- [x] `just build` — 3670 passed, 0 failed

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -879,7 +879,7 @@ markspec profile add <spec>
 
 ```sh
 markspec profile add ./profiles/my-profile
-markspec profile add @driftsys/iso26262
+markspec profile add @acme/compliance-profile
 ```
 
 #### profile publish
--- a/docs/guide/editor-vscode.md
+++ b/docs/guide/editor-vscode.md
@@ -166,7 +166,9 @@ markspec lsp install --editor neovim --binary-path /opt/markspec/bin/markspec
 
 - Confirm the binary is on `PATH`: `markspec --version` in a terminal.
 - Check the MarkSpec output panel (**View → Output → MarkSpec**) for LSP errors.
-- Verify the project has a `project.yaml` — `markspec check` requires one.
+- Confirm the file has MarkSpec entry blocks — diagnostics only appear for files
+  the server recognises (`.md` and supported source files with entry markers or
+  trace attributes).
 
 **"markspec: command not found"**
 
--- a/docs/guide/quickstart.md
+++ b/docs/guide/quickstart.md
@@ -37,18 +37,19 @@ markspec --version
 
 ## Step 2 — Create a project (1 min)
 
-> **Coming soon: `markspec init`**
->
-> A future release will scaffold a project with a single command. For now,
-> create the files by hand — it takes 30 seconds.
-
-Create a project directory and two configuration files:
-
 ```sh
 mkdir my-project && cd my-project
+markspec init
 ```
 
-Create `project.yaml`:
+`markspec init` writes `project.yaml` (minimal project metadata),
+`.markspec.yaml` (activates the bundled default profile), and `markspec.lock`,
+plus editor/agent config for any detected clients. See
+[AI agents and skillset](ai-agents.md#setting-up-a-new-project) for the full
+list of generated files and client-targeting flags.
+
+**Prefer to set it up by hand?** Two files are all you need. Create
+`project.yaml`:
 
 ```yaml
 name: my-project
--- a/docs/guide/recipes/ci.md
+++ b/docs/guide/recipes/ci.md
@@ -109,11 +109,14 @@ Generate a coverage or traceability matrix and upload it as an artifact:
 
 ## Caching the binary
 
-Cache `~/.local/bin/markspec` between runs to avoid downloading on every job:
+Cache `~/.local/bin/markspec` between runs to avoid downloading on every job.
+Bump the version in the cache key whenever you bump `MARKSPEC_VERSION` (or the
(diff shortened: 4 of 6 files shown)


stasson:
## `/review` — medium effort, 3 finder angles + verify

**What this PR does:** fixes six stale/fictional facts across the published guide + model spec book, independently re-verifying each claim against current source since the original draft was 170+ commits stale.

**Findings — 2 real, both handled:**

1. **Fixed in `ef3d2d7`**: `iso26262.md`'s new "ASIL labels" text overclaimed that `profile.labels:` enables both "completion and semantic highlighting" — only semantic highlighting is real (`lsp/semantic_tokens.ts`, `lsp/entry_ranges.ts`). No completion trigger exists for label *values* (`lsp/completions.ts` only offers the generic `Labels` trailer-*key* completion). Dropped the completion claim.
2. **Filed as #764, out of scope for this PR**: `docs/spec/model/profiles.md` uses the same fictional `@markspec/compliance-iso26262` / `@markspec/compliance-aspice` profiles as its running example throughout the *entire* document — a manifest example, an extends-chain diagram, a `.markspec.yaml` activation example, `markspec profile show` sample output, and a "Bundled profiles" table asserting they ship today. This directly contradicts what this PR just fixed in the guide book, but rewriting the model spec's illustrative example is a much bigger job than this PR's scope — filed separately.

**One candidate refuted:** a finder flagged `docs/guide/cli.md:786` as still showing `@driftsys/iso26262` — that was a stale read of `main` before this PR's branch; the diff already fixes that exact line to `@acme/compliance-profile`. Confirmed via `git show origin/docs/published-accuracy-sweep:docs/guide/cli.md`.

Everything else checked out: `markspec-schema: "1"` and the `traceability: { Satisfies: { target: [...] } }` shape are real and correctly used, `ai-agents.md#setting-up-a-new-project` and `../profiles.md` both resolve within their own book, and the discipline-mode inference algorithm matches `core/profile/discipline_mode.ts`'s actual `inferDisciplineMode` exactly.

stasson:
## Review

Fact-checked every claim in this diff against current source (not just the diff itself) — `deno.json` version, other stale references left in the docs tree, `MSL-L010`'s real definition, `ProjectConfig.labels` vs `EffectiveProfile.labels`, `inferDisciplineMode`'s actual algorithm, and `markspec init`'s actual scaffolded files. Everything holds up except one spot.

**`docs/guide/editor-vscode.md` (~line 168) — the new troubleshooting text is itself inaccurate.**

It replaces the old "`markspec check` requires a `project.yaml`" line with:

> Confirm the file has MarkSpec entry blocks — diagnostics only appear for files the server recognises (`.md` and supported source files with entry markers or trace attributes).

This conflates two unrelated LSP gates:
- `isMarkspecFile` (`packages/markspec/lsp/context.ts`) is purely extension-based (`.md` or a supported source extension) and is the *only* gate for whether a file's diagnostics get published (`server.ts`'s `documents.onDidChangeContent`/`onDidOpen`).
- `isDocCommentContext` (the "entry markers or trace attributes" heuristic) is a separate, position-level guard used only inside interactive handlers (hover, completion, definition, references, rename, document-highlight) — and only for source files, never `.md`. It has no bearing on whether diagnostics appear at all.

So the new text tells a troubleshooting reader that a recognized file needs nearby markers to get diagnostics — false for `.md` files (no such requirement) and misleading for source files (diagnostics are extension-gated, not proximity-gated; only the interactive features are proximity-gated).

Everything else in this PR checked out as accurate, including the `iso26262.md` rewrite, the `MSL-L010` correction, the `project.yaml` → `profile.labels:` swap (confirmed `ProjectConfig.labels` is a dead field beyond schema validation, while `EffectiveProfile.labels` genuinely drives semantic highlighting), the discipline-mode inference write-up, the `markspec init` file list, and the `ci.md` version bump.

---

Issue #761, opened 2026-07-04
markspec book build never renders <img> tags — markdown images silently become caption-only text


stasson:
## Summary

`markspec book build`'s renderer (`book/site/mod.ts`'s caption-detection logic, invoked via `renderChapterHtml`) treats every bare `![alt](src)` markdown image as a Figure-caption construct: it emits `<p class="caption">Figure N: alt</p>` and discards `src` entirely. No `<img>` tag is ever emitted for a standalone image, with or without a following italic caption line.

Repro:

```md
# Page

![A diagram](diagram.svg)
```

```sh
markspec book build
cat _site/page.html
```

produces:

```html
<h1>Page</h1>
<p class="caption">Figure 1: A diagram</p>
```

— no `<img>`, and the referenced file is never copied to the output directory either (a related but secondary gap, since there's no `src` for it to preserve in the first place).

## Impact

Discovered while adapting #77's mdbook→native cutover (see #189 and the follow-up branch). `docs/spec/model/index.md` embeds `![Two-layer architecture...](./architecture-layers.svg)` — under mdbook this renders as a real inline diagram; under `markspec book build` it silently becomes caption-only text with no image at all. Cutting CI over to `markspec book build` without fixing this would be a visual regression on the published Model Reference book (and any future book content that embeds a diagram).

## Suggested fix

`book/site/mod.ts`'s caption/figure detection needs to emit both the `<img src="...">` and the caption paragraph, not just the caption. The image-copy side (resolving `src` relative to the chapter's own directory and copying it into the flattened output directory, since chapters end up siblings in one `--output` dir) still needs implementing once the renderer actually preserves `src` — worth doing in the same pass.

## Not in scope here

This is unrelated to the index.md/nav-collision content-loss bug (a chapter mapped from `index.md` getting clobbered by the auto-generated nav-only `index.html`), which is fixed separately in the #77 cutover PR.

PR #762, opened 2026-07-04
feat(book): replace mdBook with native markspec book build (#77)

### LINKED ISSUE #77: Replace mdBook with native book renderer
Epic: #72

## User story

As a project maintainer, I want to remove the mdBook external dependency and use the native MarkSpec book renderer so that the CI pipeline and local builds only depend on `markspec`.

## Acceptance criteria

- [ ] `markspec book build` produces equivalent output to current mdBook builds
- [ ] Remove `book.toml` files (replaced by MarkSpec-native config in `project.yaml` or SUMMARY.md)
- [ ] Update CI workflow (`pages.yaml`) to use `markspec book build` instead of `mdbook build`
- [ ] Update `justfile` book tasks to use `markspec` commands
- [ ] No regression in published Pages site

stasson:
Closes #77.

## Summary

Adapts #189 (closed without merging) onto current `main` — the CI/justfile cutover to `markspec book build` never actually landed despite #77 being marked complete, so `justfile` and `.github/workflows/pages.yaml` were still invoking `mdbook` directly. While verifying the cutover still works, I found and fixed a real content-loss bug it would otherwise have shipped with.

- **Bug fix (`cli/commands/book.ts`)**: a chapter mapped from e.g. `index.md` was silently clobbered by the auto-generated nav-only `index.html` written right after it. Confirmed this is live today, not hypothetical: `docs/spec/model/index.md`'s entire "Overview" content (architecture diagram section, key concepts, reading-guide table) gets discarded on every `book build`. Fixed so a chapter that already claims the `index` slug wins; the nav-only page is only synthesized when no chapter does.
- **CI + justfile**: drop the mdBook install step and all four `mdbook build` invocations; build each book (`spec/language`, `spec/typography`, `spec/model`, `guide`) via `markspec book build --output <dir>` instead. Each book's output also gets its own `markspec.css` copy — `book build`'s HTML shell links `markspec.css` as a bare relative href (co-located), unlike mdbook's per-book `additional-css` config.
- **Delete** the four now-orphaned `book.toml` files (mdbook config, unused by the native renderer).
- **`docs/index.html`**: replace the FontAwesome-dependent GitHub icon — mdbook bundled FontAwesome into every book's build output, an asset the native renderer never produces — with an inline SVG, dropping the dependency entirely.
- **Remove `book-dev`** justfile target (`mdbook serve` has no direct replacement; `markspec book dev` is a registered-but-unimplemented CLI stub, per `AGENTS.md`'s own "Not yet implemented" table).

A separate, pre-existing gap — markdown images never render as `` tags at all (swallowed into caption-only text by the renderer's figure-caption detection) — is real and would affect `docs/spec/model/index.md`'s embedded diagram, but is unrelated to this cutover and a materially bigger fix (touches `book/site/mod.ts`'s caption-detection AST walker). Filed separately as #761 rather than expanding this change.

## Test plan

- [x] New e2e test: a chapter mapped from `index.md` renders its own content in `index.html`, not the auto-generated nav (reproduces the bug before the fix, passes after)
- [x] Ran `just book` locally end-to-end — all 4 books build via `markspec book build`, no mdbook involved
- [x] Verified every book's internal nav links resolve, `markspec.css` is present alongside each book's HTML, and the root `docs/index.html`'s links (`guide/`, `spec/`, `model/`, `typography/`, `markspec-cheatsheet.pdf`, `theme/markspec.css`) all resolve to real output paths
- [x] Confirmed the FontAwesome removal — `_site/index.html` has zero references to it and renders the inline SVG instead
- [x] `just build` — 3671 passed, 0 failed


stasson:
## `/review` — medium effort, 4 finder angles + verify

**What this PR does:** cuts the CI/justfile book build over from mdBook to native `markspec book build`, fixing a content-loss bug (a chapter mapped from `index.md` was silently clobbered by the auto-generated nav-only `index.html`) discovered while verifying the cutover.

**Findings — 2 real, both fixed in `3130bf5`:**

1. **The index.md fix itself introduced a regression**: skipping the nav-only `index.html` whenever a real chapter claims that slug strands every *other* chapter with no site-wide navigation. Verified against the actual model book: `prose-analysis.html` and `annex-profile-schema.html` (2 of 8 chapters) aren't mentioned in `index.md`'s own hand-written "Reading guide" table, so they'd have shipped completely unreachable from the homepage — trading the original content-loss bug for a navigation-loss bug. Fixed by always appending the full chapter nav section to whichever page becomes `index.html` (the real chapter's content, or the synthesized page) — content and navigation are no longer either/or. Strengthened the e2e test to assert every chapter stays linked, not just the index-mapped one, and re-verified against the real model book (all 8 chapters + Overview content present).
2. **`schemas/index.html`** — a sibling landing page the original fix missed — had the exact same FontAwesome-dependent GitHub icon as `docs/index.html` (both depended on mdBook bundling FontAwesome into `_site/spec/`, an asset the native renderer never produces). Applied the identical inline-SVG fix.

**Also addressed:** extracted the slug-computation formula (duplicated across two call sites this diff already touched) into a `_slugFor` helper, and replaced the mutable `hasIndexChapter` flag with a pure lookup.

**Non-blocking, left as-is:**
- The four-book list (`spec typography model guide`) and its `deno run .../book build --output ...` invocation is hand-duplicated 4× in `pages.yaml` and again 4× in `justfile`, even though this same diff proves a `for book in ...` loop works for the CSS-copy step two lines below. A shared list would remove the drift risk of a future book edit updating one but not the other. Didn't restructure this now given the two files are different execution environments (GH Actions steps vs `just` recipes) with no trivially shared loop construct across both.
- `theme/markspec.css`'s mdBook-theme override block (`.light, .rust`, `.navy, .ayu, .coal` selectors) is now dead CSS — the native renderer's HTML never sets those classes. Harmless, but worth pruning in a follow-up.
- `docs/spec/internal/markspec-user-docs.md` (an unpublished internal draft, not in any `SUMMARY.md`) still describes the now-removed mdBook pipeline in several places — docs-staleness only, no runtime impact.

stasson:
## Review

This PR fixes a real content-clobbering bug (an `index.md`-mapped chapter getting overwritten by an auto-generated nav page) and cuts CI/justfile over from mdBook to the native `markspec book build` renderer. Verified findings below by reading the diff, `git show`-ing full files on the PR branch, and cross-checking claims against the real doc content in this repo (not just the diff).

### Blocking — verified regressions in the exact content this PR's fix targets

**1. Cross-chapter Markdown links are never rewritten from `.md` to `.html`.**
`packages/markspec/book/site/mod.ts` runs a plain remark → rehype pipeline (`remarkParse` → `remarkGfm` → `remarkRehype` → `rehypeStringify`) with no link-rewriting step. `docs/spec/model/index.md:58-63` — the exact chapter whose content-loss this PR fixes — links `[Entry format](entry-format.md)`, `[Type taxonomy](types.md)`, etc. Once built, the anchor still points at `entry-format.md`, but `_slugFor` names the actual output file `entry-format.html`. These links 404 in the published site.

**2. Nested chapters are flattened into one output directory, but their relative links aren't adjusted for it.**
`docs/guide/recipes/shipping-a-reference-architecture.md:13` links `[Profile specifiers](../profiles.md#profile-specifiers)`, intending to reach its sibling `docs/guide/profiles.md`. `_slugFor`'s flattening (`recipes/shipping-a-reference-architecture.md` → `recipes-shipping-a-reference-architecture.html`) puts that chapter in the same directory as `profiles.html`, so the literal `../profiles.md` now resolves *outside* `_site/guide/` entirely — a second, compounding way real content breaks under the new pipeline.

### Non-blocking

**3. Per-book title/branding is lost on synthesized homepages.**
Books without an `index.md` chapter (spec/language, spec/typography, guide) now render `<title>io.driftsys.markspec</title>` / `<h1>io.driftsys.markspec</h1>` (from `project.yaml`'s `name:`) instead of the deleted `book.toml` titles (e.g. "MarkSpec Language Specification"). The model book's homepage similarly shows the bare chapter title "Overview" instead of "MarkSpec Model Reference".

**4. The clobbering fix only special-cases the literal `"index"` slug, not the general collision class.**
Two chapters whose paths flatten to the same slug (e.g. `a/b.md` and `a-b.md`) still silently overwrite each other with no warning — the same failure class this PR sets out to fix, just untriggered by any current book's content.

**5. The 4-book list is hardcoded independently in 4 places** (2 build-step lists + 2 CSS-copy loops, split across `.github/workflows/pages.yaml` and `justfile`). Adding/renaming a book means editing all 4 spots; missing the copy-loop entry ships an unstyled page with nothing in CI to catch it.

**6. Redundant slug computation** — `chapterLinks` computes each chapter's slug once via `.map()` (book.ts), then the write loop calls `_slugFor(chapter.path)` again for the same chapter instead of reusing `chapterLinks[i].slug`.

**7. `_escHtml` (new, in `book.ts`) doesn't escape `"`, while the pre-existing `_escapeHtml` in `book/site/mod.ts` does the same job and escapes it.** Not exploitable today (both call sites place it in text-node position, never inside an attribute), but a latent trap for the next caller that does.

### Already known, not re-flagged

Missing `` rendering and missing asset-copy for referenced images/SVGs — already filed by this PR's author as #761 and explicitly called out in the PR description as out of scope here.

---

PR #765, opened 2026-07-04
feat(core): federated upstream slice 4 — graph integration + validator

### LINKED ISSUE #744: story(core): graph integration + validator (federated upstream slice 4)
Part of #741 (slice 4 of 6). Depends on slice 1 (shipped, `60ceec5`) and #742
(slice 2 — manifest + lockfile rows).
Design: `docs/wip/2026-07-04-federated-upstream-resolution-design.md` §4.5,
§4.6, §4.7, §4.9.

## Goal

Hydrate locked upstream snapshots into the compiled graph as read-only
citizens and wire the validator to resolve, classify, and gate on them.

## Scope

- Three feed sites, in declaration order (dependencies, then references),
  before project files: CLI compiler (`compileProject`, immediately after
  delivered corpus), LSP server (`seedUpstreamCorpus()` beside
  `seedDeliveredCorpus()`, re-seeded on lockfile/cache change), MCP server
  (via the shared compile cache).
- `deserializeCompileResult` (inverse of `serializeCompileResult`) guarded
  by a schema-skew check — a snapshot whose `markspecSchemaVersion` /
  `generator.coreSchema` doesn't match is rejected with a "re-lock with a
  compatible markspec version" diagnostic, never a silent misparse.
- `loadUpstreamCorpus` (built in slice 1) wired to the lockfile: reads the
  cached snapshot for each locked upstream, hydrates `Entry[]`, stamps
  `origin = { kind: "upstream", upstreamId, version }`.
- **Authoritative-source rule:** hydration skips snapshot entries that
  already carry an `origin`, so a root pulling both A and B (where A's
  snapshot also contains B's re-exported entries) only takes B's entries
  from B's own snapshot — this is what makes the diamond/root aggregation
  pattern collision-free.
- Flat resolution: trace targets resolve against project + corpus +
  dependencies + references. `MSL-L006` / broken-ref diagnostics stop firing
  for IDs found upstream.
- **`MSL-T014` lands:** when any dependency/reference is declared,
  unresolved-after-chain refs fire T014 instead of L006, naming the searched
  set (e.g. "not found in project or upstreams: product, icd, refhub").
- Coverage semantics differ by field: `dependencies:` participate in
  coverage analysis (a product requirement with no component coverage is a
  reported gap); `references:` are leaves, never expected to be covered or
  to cover.
- Generalize the ADR-030 collision pass: project↔upstream, upstream↔upstream,
  upstream↔corpus all produce the `MSL-R014` shape naming both origins.
  No precedence — internal first-entry-wins (declaration order) only keeps
  the graph functional while the diagnostic fires.
- Upstream entries are validation-exempt, not edge-inert: no prose lint, no
  structural checks, refs *from* upstream entries never produce
  unresolved-ref diagnostics — but when such a ref resolves inside the flat
  union (e.g. component A satisfying component B, both hydrated at a root),
  the edge joins the graph like any other.
- Unreleased-pin advisory (design §4.4): a branch/sha-resolved dependency
  produces a non-blocking project-level advisory below `--strict`; under
  `check --strict` it fails, naming the dependency and the remedy (cut a
  tag, then `lock --update`).
- Root/program pattern (design §4.9) falls out of the above once flat
  resolution + coverage semantics are correct: verify program-wide
  `report`, `dependents`, and `context` work at a root with cross-component
  edges.

## Out of scope

- LSP/MCP-specific UI surfaces beyond feed-site wiring — completion badges,
  hover, definition no-op (slice 5).
- Diagnostic-code assignment for the unreleased advisory beyond "some
  code" — align with the ADR-012 phased catalogue at implementation time
  (design §10, open question).

## Testing

- Unit: hydration round-trip (`deserialize(serialize(x))` ≍ `x` modulo
  origin); collision-pass tables; schema-skew rejection.
- E2E: broken ID → `MSL-T014`; colliding ID → `MSL-R014`; deleted cache →
  drift gate; branch-resolved pin → advisory, then `--strict` failure.
- E2E root pattern: three temp projects (root depends on A and B; A also
  depends on B — diamond). Assert no duplicate/collision diagnostics
  (authoritative-source rule), A→B edge present in the root's matrix
  report, `dependents` at root crosses the repo boundary, coverage report
  flags a product requirement no component satisfies, and the root's
  published `/api/` entry count equals A-authored + B-authored +
  root-authored.
- Snapshots: diagnostic wording for T014, collision, drift, unreleased
  advisory.

stasson:
Closes #744.

Slice 4 of the federated-upstream epic (#741) — the payoff slice: locked
upstream entries now **resolve in the compiled graph**, so a downstream
`Satisfies:` (or any trace link) to an entry authored in another repository
resolves and validates. Design:
`docs/wip/2026-07-04-federated-upstream-resolution-design.md` (§4.6 feed sites,
§4.7 validator behaviour, §4.8 surfaces, §4.9 root/diamond). Executed against
`docs/wip/2026-07-04-federated-upstream-slice4-plan.md` (TDD per task, reviewed,
then a whole-branch review).

## What ships

**Four feed sites.** The already-cached upstream entries (slice 1's
`loadUpstreamCorpus`, slice 2's `[[upstream.registry]]` rows + cache) are fed
into the graph at every site, injected ahead of project entries like the
ADR-030 delivered corpus, all soft-failing on a cold cache:

- **CLI compiler** — `compileProject` (covers `compile`, `export`, `show`,
  `context`, `dependents`, `report`).
- **`check`** — the `runPipeline` path (lockfile parse hoisted, drift gates
  reuse it).
- **LSP** — `seedUpstreamCorpus()` beside `seedDeliveredCorpus()`, re-seeded on
  `markspec.lock` change; upstream files join the read-only guard set.
- **MCP** — `runCompile` reads the lockfile, feeds upstream entries; `isStale`
  invalidates on a re-lock.

**Validator behaviour.**

- **Flat resolution** — trace targets resolve against project + corpus +
  upstreams; `MSL-L006` stops firing for IDs found upstream.
- **`MSL-T014` lands** (warning) — when a project declares
  `dependencies:`/`references:`, an unresolved-after-federation ref fires T014
  naming the searched set, instead of the generic L006. (language.md §8.3
  reservation → active row.)
- **`MSL-R014` generalized** — a project ID colliding with an upstream-provided
  ID is an R014 error; the message names the origin(s) generically (project↔
  upstream, upstream↔upstream, upstream↔corpus).
- **Validation-exempt, edge-live** — `kind:"upstream"` entries are skipped from
  every per-entry emitting check (structural, T005/T012, classification,
  trace-type, discipline, late-stage inference, typed-attribute, trace, and
  typl) while remaining resolution targets; their `type` comes from their own
  compile (never re-classified). Delivered corpus (`kind:"profile"`) is
  unchanged.
- **Coverage** — `references:` entries are traceability leaves (never coverage
  gaps); `dependencies:` participate.

**Root/program pattern (§4.9).** A root repo declaring member repos aggregates
the program graph — proven end-to-end by a blackbox diamond e2e (ROOT←B, ROOT←C,
C←B): the authoritative-source rule counts B's entries exactly once (via B's
direct reference, not C's re-export), the C→B cross-repo edge resolves at ROOT,
and no collision fires.

## New public API

`upstreamRefsFromLockfile(lockfile, projectRoot)` (pure row→ref mapping),
`loadProjectUpstreams` (CLI loader), `isUpstreamEntry(entry)` predicate.

## Scope boundaries

- **References resolve; git dependencies do not yet.** Git dependency
  *acquisition* is slice 3 — the row→ref mapping and coverage classification
  already accept `[[upstream.dependency]]` rows so slice 3 drops in, but no
  dependency entries hydrate today. The §4.4 unreleased-pin advisory +
  `check --strict` pin gate is deferred to slice 3 (needs dependency rows).
- **LSP/MCP user-facing surfaces** (completion badges, hover, definition no-op,
  decorations) are slice 5 — this slice makes upstream entries present,
  resolving, and read-only at those sites.
- The ADR + `language.md` §3.2/compile-output rewrites are slice 6; only the
  user guide is updated here.

## Verification

`just build` green: **3765 passed, 0 failed, 2 ignored**; `deno fmt --check`
and `dprint check` clean. A whole-branch review returned ready-to-merge after
two fixes it caught (upstream exemption extended to `validateTypl`; MCP
lockfile-mtime committed on compile success only). The e2e suite includes the
6-scenario cross-repo resolution suite (`federated_resolve_test.ts`) — Satisfies
resolves, broken ID → T014, colliding ID → R014, reference is a coverage leaf,
origin surfaces in `show`/`report`, and the root/diamond dedup.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -94,6 +94,53 @@ does not constrain which `Labels:` values entries may carry.
   tag or branch name; version-selecting resolution is planned for a future
   release.
 
+#### Upstream entries resolve in the graph
+
+Once `markspec lock` has pinned a `references:` entry, its entries join this
+project's own traceability graph as read-only, origin-tagged citizens — not just
+cached files on disk:
+
+- **Trace links resolve across the repo boundary.** A `Satisfies:` (or any other
+  trace attribute) value that names an upstream display ID resolves exactly like
+  a same-project reference. `check`, `show`, `context`, `dependents`, and
+  `report` all see the upstream entry, tagged with an `Origin:` line / column
+  showing `<upstreamId>@<version>` — see [`show`](#show) and [`report`](#report)
+  below.
+- **`MSL-T014` replaces `MSL-L006`** for a trace value that still doesn't
+  resolve once the project declares any `dependencies:`/`references:`: the
+  warning names every upstream searched, e.g.
+  `not found in project or
+  upstreams: producta, icd`. A project with no
+  declared upstreams keeps the plain `MSL-L006` behavior.
+- **A project entry that reuses an upstream's display ID or `Id:` fails `check`
+  with `MSL-R014`**, naming the colliding upstream origin — the same shape as
+  the collision that fires when a profile
+  [delivers a corpus](profiles.md#delivered-documents) (ADR-030), generalized to
+  cover upstream origins too. Fix it by renaming the project entry; upstream
+  entries are read-only.
+- **`references:` entries are traceability leaves.** `report coverage` never
+  reports one as an orphan or unsatisfied gap — a citation isn't something your
+  own project is expected to cover. `dependencies:` entries participate in
+  coverage like a project entry — but since git dependency acquisition hasn't
+  shipped yet (above), no `dependencies:` entry hydrates into the graph today,
+  so this only takes effect once that lands.
+- **Upstream entries are validation-exempt, not edge-inert.** No structural
+  checks or prose lint run against them — that already happened in their own
+  repo's `check` — but they remain full resolution targets: a project entry's
+  trace link to one still resolves, and so does an edge between two upstream
+  entries once both are hydrated into the same graph.
+
+**Root/program project pattern.** A root or program repository that references
+every member repo aggregates the whole program's graph for free: each member's
+entries hydrate into the root's compile, cross-repo trace edges resolve there,
+and `report`, `dependents`, and `context` run against the entire program from
+the root — both ends of every cross-repo edge are present. A repo referenced
+through more than one path (a diamond — e.g. the root references both a
+component and something that itself references that component) is still counted
+exactly once: an upstream snapshot's own re-exported entries are skipped in
+favor of that entry's authoring repo, so aggregating never double-counts or
+collides.
+
 ### .markspec.yaml
 
 `.markspec.yaml` carries markspec's own tool configuration: profile binding
@@ -271,7 +318,10 @@ reference values:
   `markspec.lock` records the stable `target-ulid`; `fmt` uses it to rewrite the
   stale display ID to the target's new name.
 - **Unresolved references are left as-is** — `markspec check` reports them via
-  MSL-L006.
+  `MSL-L006`, or `MSL-T014` when the project declares `dependencies:`/
+  `references:` (see
+  [Upstream entries resolve in the graph](#upstream-entries-resolve-in-the-graph)
+  above).
 
 Neither action is performed when no `project.yaml` is discoverable (file-local
 invocation). `fmt` reads the edge ledger but never writes `markspec.lock`; the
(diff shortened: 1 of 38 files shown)


stasson:
## Pre-merge review (whole-branch, 14 commits)

An independent whole-branch review ran before this PR opened (in addition to the per-task reviews). It verified the four feed sites are consistent (all map lockfile→refs via the one pure `upstreamRefsFromLockfile`, inject upstream ahead of project, soft-fail on cold cache), the validation-exemption is disciplined, and the authoritative-source rule is proven end-to-end by the diamond e2e. It caught **two real issues, both fixed in `08178c5` before this PR**:

1. **Exemption gap — `validateTypl` (Important).** `validate()` (Stage 1, used by check *and* compile) called `validateTypl(entries)` over *all* entries, so a hydrated upstream entry's typl content emitted error-severity `TYPL-005/009/010/011` — contradicting the "validation-exempt" guarantee, and the consumer-side `TYPL-009` declared-once path could fail `check` on a project entry re-declaring an upstream-published dotted name. **Fix:** `validateTypl` now builds its validation registry and runs its emit loops over non-upstream entries only (upstream entries are typl-inert for validation; the compile-output `typeRegistry` for LSP completions is untouched). Added RED→GREEN tests.
2. **MCP cache coherence (Important).** `lockfileMtime` was committed before `compile()` succeeded, so a failed lock-change recompile could leave the invalidation key advanced while the cache held the pre-lock result → a later `getCompiled()` serves stale. **Fix:** moved the assignment into the commit-on-success block.

**Deferred (non-blocking) Minors**, triaged for follow-up rather than merge: no explicit `kind:"dependency"` fixture in `refs_test` (slice 3 lands dependency loading); malformed-lockfile diagnostics dropped in `compileProject` (by-design — integrity is `check`'s MSL-L212 job); missing-cache reported by two orthogonal gates (mirrors compile); a couple of test-hardening nits (URI-with-upstreams, T014 message equality); the `attributeCorpusDiagnostics` "delivered by X:" prefix is corpus-flavored on the rare upstream↔upstream downgrade; `report.ts` double-loads config; a "resolve" wording overload in the guide.

**Verification:** `just build` 3765 passed / 0 failed; `deno fmt --check` + `dprint check` clean.

🤖 Pre-merge review via [Claude Code](https://claude.com/claude-code)

stasson:
### CI green — Windows path-separator fix

`Test (windows-latest)` initially failed (macOS/Linux passed) — a Windows-only cache-path bug: `upstreamRefsFromLockfile` built the cache dir as `` `${cacheRoot}/${id}` ``, mixing `join()`'s backslash output with a literal `/`, so in-memory test fixtures keyed on exact path strings didn't match on Windows (the e2e passed because a real filesystem normalizes separators). Fixed by using `join()` for every cache-path construction — production (`refs.ts`, `loadUpstreamCorpus`, and the slice-2 `writeCache`/`probeCacheSnapshot` so writer and reader agree) and every in-memory test fixture key (`15e31f0`, `b2350c2`). All 17 checks now green.

stasson:
## `/review` (8-angle multi-agent) — 2 confirmed bugs the per-task/whole-branch reviews missed

Ran the full independent `/review` on the PR diff. Two findings verified CONFIRMED against the code; the rest are altitude/efficiency.

### 1. Cross-repo `Satisfies:` to an upstream entry of a *different type vocabulary* fails `check` — `core/validator/traceability.ts:191` (Important, CONFIRMED)
The slice-4 exemption skips upstream *sources* but not upstream *targets*. `matchesAnyTarget` does raw type-string equality with no target-side exemption and emits `MSL-L004` at **error** severity. Scenario: downstream rule `Satisfies → ["system-requirement"]`; the resolved upstream target is typed `"SystemRequirement"` (its own profile's vocabulary — D6: "the consumer never re-classifies") → mismatch → L004 error → `check` exits 1. Before slice 4 this was a benign `MSL-L006` warning (target unresolved). This regresses the feature's own use case (cross-repo refs between differently-typed repos). Only the same-vocabulary case is tested. Fix: exempt upstream *targets* from the target-type check, mirroring the source-side exemption.

### 2. Coverage counts treat reference leaves inconsistently — `core/reporter/mod.ts:234` (Minor, CONFIRMED)
`withSatisfies++` isn't gated by `!isReferenceLeaf`, while the orphan/unsatisfied branches are. A `references:` upstream entry with a resolving outgoing `Satisfies:` inflates the coverage numerator while the same leaf without one is excluded from the gap lists — the two numbers no longer describe the same population. Fix: gate the increment on `!isReferenceLeaf`.

### 3. Unfixable `MSL-R014` remedy on upstream↔upstream ID collision — `core/validator/corpus.ts` (Minor)
Two `references:` that independently author the same display ID emit an R014 error whose "rename this entry" remedy the consumer cannot perform (both read-only). The collision firing is by-design (flat namespace, D5); the remedy wording is unactionable in this branch.

### Altitude / efficiency (deferred follow-ups, not blockers)
- The validation-exemption guard is scattered across ~10 emit loops; this PR proved it fragile (`validateTypl` leaked until the whole-branch review caught it, and #1 is the same class). Partitioning `emittable` vs `resolution-only` entries once at the pipeline top would structurally prevent the next stage from re-introducing the gap.
- The four feed sites duplicate the lockfile→upstream hydration chain; `loadProjectUpstreams` should move to `core/upstream` with an injected `readFile` so LSP/MCP can share it.
- The LSP `reloadProfile` fires on `markspec.lock` changes but unconditionally re-resolves the whole profile chain + re-seeds the delivered corpus + pushes `profileChanged`.
- `report.ts` double-loads `project.yaml`.
- Declaring any `references:` rewrites all local unresolved-ref typos from L006 → T014 (spec-intended; noted for L006-grep tooling).

Fixing #1 and #2 in a follow-up commit on this branch.

🤖 `/review` (8 finder angles + verify) via [Claude Code](https://claude.com/claude-code)

PR #772, opened 2026-07-04
feat(core): typl table declaration surface (#724)

### LINKED ISSUE #724: story(typl): table surface for typl declarations
Part of #717 §C.2. Depends on: S3 (table surface), S4 (base-resolution). Thin — rides the table surface + resolver.

## Goal
Allow typl declarations in a table: `$name | kind shape | description`, one row per binding.

## Acceptance
- Table-form typl declarations produce the same bindings as fence/bullet/inline.
- Unit-test parity with existing surfaces.

stasson:
## Summary

S6 of the uxil epic (#717 §C.2): adds a **fourth typl embedding surface** — GFM
tables — on the shared declaration machinery built in S2/S3 and the
base-resolution engine from S4. Thin by design: a ~30-line adapter plus a small
wire-in to `assemble.ts`.

Each data row `$name | kind shape | description` is one binding: the first two
cells are read positionally, the description column is dropped (typl bindings
carry no description), and rows whose first cell is not a `$Name` are skipped —
so a table may mix declaration rows with plain notes. The row is reduced to a
`$name : kind shape` source and parsed through the same `parseTyplBlock` the
other surfaces use, so a table row yields the **identical `Binding`** as the
fence / bullet / inline form.

A `Table:` caption may carry an absolute base (`$a.b`); the table's relative
rows (`$.x`) resolve against that caption base first, then the entry root,
through the entry-local resolver (#722 rule 3). The caption base is not a root
declaration and never sets `rootNamespace`. Union/enum shapes containing `|` are
authored with the standard GFM `\|` escape (the cell un-escapes before typl
parses it) — resolving the concern that made ADR-019 originally reject a table
surface.

## Changes

- **`core/typl/table.ts`** (new) — `typlTableRowRecognizer`,
  `typlTableCaptionBase`, and `extractTyplTable` (a thin adapter over
  `core/decl/table.ts`), exported from `typl/mod.ts`.
- **`core/typl/assemble.ts`** — table loop (mirrors the fence loop) + caption
  base threaded into `scopeFor` as the innermost scope.
- **Tests** — recognizer + caption-base unit tests (`table_test.ts`); assemble
  parity (table ≡ inline), caption-base resolution, mixed-table, escaped-union,
  and cross-surface composition cases.
- **Docs** — ADR-019 addendum (revisits the rejected-table alternative), typl
  guide + language-reference table sections.

## Acceptance (#724)

- [x] Table-form typl declarations produce the same bindings as
      fence/bullet/inline.
- [x] Unit-test parity with existing surfaces.
- [x] Caption base wired into resolution (approved scope).

`just build` green (lint + full test suite + typecheck + compile). No
`CORE_SCHEMA_VERSION` bump.

Closes #724.

### DIFF
--- a/docs/architecture/adr-019-typl-type-dsl.md
+++ b/docs/architecture/adr-019-typl-type-dsl.md
@@ -82,3 +82,29 @@ applies. `CORE_SCHEMA_VERSION` unchanged.
 Full design record: `docs/wip/2026-07-04-typl-published-tier-design.md`
 (gardened to `docs/archive/` when this branch lands). The complete namespacing
 rewrite of this ADR plus guide chapter is story #730.
+
+## Addendum: table surface (#724, 2026-07-05)
+
+S6 of the uxil epic (#717) adds a **fourth** Markdown surface: a GFM table. Each
+data row `$name | kind shape | description` is one binding — the first two cells
+are read positionally (the description column and any further columns are
+documentation), rows whose first cell is not a `$Name` are skipped, and only
+bindings (not typedefs) are expressed this way. Like its siblings it is a thin
+adapter (`core/typl/table.ts`) over the shared declaration machinery
+(`core/decl/table.ts`, S3): the row recognizer reconstructs a
+`$name : kind
+shape` source that `parseTyplBlock` parses, so a table row yields
+the identical `Binding` as the fence, bullet, or inline form.
+
+A table's `Table:` caption may carry a published base (an absolute `$a.b` name);
+the table's relative rows (`$.x`) resolve against that caption base first, then
+the entry root, through the entry-local base-resolution engine (#722 rule 3).
+The caption base is not a root declaration and never sets `rootNamespace`.
+
+This revisits the "GFM bindings table … rejected" alternative above. Two things
+changed since 2026-05-25: a shape carrying `|` (a union or enum) is authored
+with the standard GFM `\|` escape — the cell un-escapes before typl parses it —
+and ADR-029's whole-document dprint pass preserves table line structure. The
+table is therefore offered as an _additional_ surface, not the primary one;
+dense or `|`-heavy declarations still read best in a fence.
+`CORE_SCHEMA_VERSION` unchanged.
--- a/docs/guide/typl.md
+++ b/docs/guide/typl.md
@@ -99,7 +99,7 @@ table may mix declaration rows with plain notes.
   | $VehicleSpeed | signal float[0..300] | km/h, road speed |
   | $EngineRpm    | signal int[0..8000]  | crankshaft speed |
 
-      Id: 01JZEXAMPLEULID000000000004
+      Id: 01JZEXAMPLEULID000000000006
 ```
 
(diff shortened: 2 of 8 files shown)


stasson:
## Automated review (`/review`, medium effort — 8 finder angles → verify)

Reviewed the diff for correctness, cleanup, altitude, and CLAUDE.md conventions. One real correctness bug surfaced plus a few hygiene/accuracy items; all actionable ones are fixed in the follow-up commit `9ae1048`.

### Fixed

1. **Correctness — table-row diagnostic reported one line too low.** `assemble.ts` computed the diagnostic offset as `bodyStartLine + row.range.start.line - 1` (the *fence* convention). But a table row's range points **at** the row, like the bullet surface (whose range also points at content and uses `-2`), whereas the fence range points at the opening ` ``` ` a line above its content. So a malformed / unresolved table row (e.g. a relative `$.x` with no base → `TYPL-010`, or a row that fails to parse) had its diagnostic anchored one line below the row. Fixed `-1` → `-2` and added a regression test asserting the diagnostic lands on the row's own line. *(The five acceptance tests were all zero-diagnostic happy paths, so this slipped through.)*

2. **Convention — over-export at the `typl` library boundary.** `mod.ts` re-exported `typlTableCaptionBase` and `typlTableRowRecognizer` alongside `extractTyplTable` + its type. The sibling surfaces (`fence`/`bullet`/`inline`) export only their `extract*` function + extraction type — their recognizer stays internal to `recognize.ts`. No consumer imports the two helpers via `mod.ts` (`assemble.ts` and the colocated test import them from `./table.ts` directly), so they've been dropped from the re-export to match the convention and keep the boundary narrow.

3. **Doc accuracy — nested-table scope.** The `scopeFor` doc comment read as if a table row walks a bullet-ancestor chain; a table row carries no `bulletParent`, so a table nested inside a bullet-namespace subtree sees only its caption base + the entry root, never the enclosing bullet base. Corrected the comment and documented this as a known limitation in both the guide and the language reference (the docs previously implied full parity with the bullet surface's nested namespaces).

4. **Docs — duplicate example `Id`.** The new guide table example reused `Id: …004`, colliding with the pre-existing "Mixing surfaces" example. Bumped to a unique value.

5. **Docs — single-segment caption base.** The caption base was described only as an absolute `$a.b` name, but the code also accepts a single-segment `$vehicle`. Noted the single-segment form in the guide + reference.

### Considered, no change

- **`namespace` table row becoming an entry root** — not a defect: a `$foo : namespace` binding is treated identically across all four surfaces, so the table behaves consistently. The code comment there refers specifically to the *caption base* (which correctly takes no part in root detection) and is accurate.
- **Caption base dropped when punctuation abuts the name with no space** (`$a.b—desc`) — low; the documented convention uses a spaced em-dash, and the first-token parse is intentional.
- **`ABSOLUTE_TYPL_NAME_RE` as a fourth name-grammar copy** — kept: it's a deliberate, documented single-segment variant of `isPublishedTyplName` (which requires ≥2 segments); centralizing it would add another exported predicate for one call site.

Full `just check` green after the fixes (**3670 passed, 0 failed**).

PR #784, opened 2026-07-05
feat: federated upstream slice 3 — git dependency fetcher (#743)

### LINKED ISSUE #743: story(core): git dependency fetcher (federated upstream slice 3)
Part of #741 (slice 3 of 6). Depends on slice 2 (manifest loader + lockfile
rows + lock flows).
Design: `docs/wip/2026-07-04-federated-upstream-resolution-design.md` §4.3,
§4.4.

## Goal

Acquire and compile `dependencies:` git repositories inside `markspec lock`,
bounded by lock events rather than daily work — `check`/`compile`/LSP only
ever read the resulting cached snapshot.

## Scope

- **Never clone to resolve.** Version-intent resolution is `git ls-remote`
  only (tags + heads); auto/tag intents semver-sort the tag list (default
  pattern `v*`).
- **Never fetch history.** Acquire the tree at one sha: prefer the forge
  tarball endpoint for a recognized GitHub/GitLab host (one authed API
  request, no `.git` directory); else fall back to
  `git clone --depth 1 --filter=blob:none` at the sha.
- **Content-addressed cache.** Snapshots are immutable per sha. Compile the
  acquired tree in-process (`compileProject`), producing the same JSON a
  published reference would serve.
- Record the resolution kind (`tag` \| `branch` \| `sha`) into the slice-2
  lockfile row so `check --strict` (slice 4) can enforce pin-level release
  assurance.
- v1 constraint: the acquired tree must be self-contained enough to
  compile — its own profile chain must resolve from its own tree (a
  profile-fetch during `lock` is acceptable; `lock` is the one designated
  online step).
- Document the CI recipe: cache `.markspec/cache/upstreams/` keyed on the
  lockfile hash, so CI only pays acquisition cost when a pin actually moved.

## Out of scope

- Owner-publishes escalation (a dependency served from a published site
  instead of cloned — design §4.3, reuses the slice-2 reference fetcher,
  returns the deferred `kind` discriminator when it lands).
- Shared remote snapshot cache (org-level, keyed `(repo, sha)`) — deferred,
  needs infrastructure (design §4.3, ADR-020 trajectory).
- Per-file forge-API crawling — rejected outright (design §8): the file set
  isn't known without discovery, so it degenerates into many round-trips
  that lose to one tarball.

## Testing

- Unit: semver tag sort; ls-remote intent resolution against fixture ref
  lists.
- E2E: a local bare-repo fixture with tags as a `dependencies:` entry (no
  network) — assert `lock` pins to the correct sha for `auto`/tag/branch
  intents, and that a second `lock` run does not re-acquire an unmoved pin.

stasson:
Closes #743. Part of #741 (federated upstream, slice 3 of 6).

## What this ships

The **producer** half of git dependencies: `markspec lock` now acquires each
`dependencies:` git repository and compiles it into a cached snapshot, so
`check`/`compile`/LSP resolve cross-repo `Satisfies:` traces offline. Slices 2
and 4 already shipped the entire *consumption* side (hydration, graph
integration, `MSL-L212` drift, the `UpstreamDependency` row model), so this
slice is purely the producer — everything downstream just works once the row
and cache exist.

The acquisition ladder, per the design (§4.3):

1. **Resolve intent** with `git ls-remote --symref` only (tags + heads) —
   `auto` picks the highest semver tag (else the default-branch head); an
   explicit `<tag>`/`<branch>`/`<sha>` resolves directly.
2. **Acquire the tree at one sha** via a shallow, hermetic
   `git fetch --depth 1 --filter=blob:none` + `checkout FETCH_HEAD` (no history,
   no `.git` left behind).
3. **Compile in-process** into the same `manifest.json` + `compiled.json`
   layout a published reference serves, cached under
   `.markspec/cache/upstreams/<id>/`, and record a `[[upstream.dependency]]`
   row (`intent` / `resolved` kind / `sha` / `snapshot` / `locked-at`).

Three lock flows: **first-lock**, **keep/restore** (re-acquire the *pinned* sha
only to repopulate a missing cache — never re-resolve), and **update**
(`--update` moves the pin).

## Design decisions settled here

1. **Warn-and-write (unified).** An unreachable `references:` *or*
   `dependencies:` upstream now emits a **warning** and `lock` still writes
   every pin that resolved — reversing slice 2's all-or-nothing abort. A
   transient 404 or offline run no longer drops the whole lockfile.
2. **Clone-only in v1.** Acquisition is shallow `git fetch`-by-sha; the
   GitHub/GitLab forge-tarball optimization is deferred to a fast-follow
   (it needs an authed host and can't be exercised by the offline e2e).
3. **`check --strict` release gate.** A new `MSL-L215` advisory fires once per
   dependency pinned to a branch/sha (an unreleased state); below `--strict`
   it's a gentle warning, `--strict` promotes it to a failing error — reusing
   check's existing warning→error promotion (no bespoke gate).

## Determinism (a correctness invariant the epic design didn't cover)

A dependency is compiled **locally**, so its `compiled.json` — whose sha256 is
the `snapshot` pin shared across machines in `markspec.lock` — must be
byte-reproducible from `(source sha, markspec version)`. Achieved by compiling
with **sorted, tree-relative paths** and **omitting the stat/git callbacks**, so
no absolute path, mtime, or git history leaks into the output. The unit test
guards this against a dropped sort using a nested fixture where depth-first
order ≠ lexicographic order.

## Security & robustness (from the final review + the pre-push gate)

- **Git-URL transport-helper RCE — hardened.** `dependencies[].url` comes from
  an untrusted `project.yaml`; a URL like `ext::sh -c '…'` or a leading-`-`
  `--upload-pack=…` could run code at `lock` time. Closed with
  `GIT_ALLOW_PROTOCOL=https:ssh:git:file` on the git subprocess **and** a guard
  rejecting empty/leading-dash URLs (which flow to the warn-and-write path).
  Verified: `ext::` now hits `fatal: transport 'ext' not allowed`, the command
  never runs.
- **Hermetic git subprocess.** `markspec lock`'s git calls now ignore an
  ambient `GIT_DIR`/`GIT_WORK_TREE`/… so acquisition works correctly when
  `lock` is invoked from within a git hook or a `GIT_DIR`-setting CI (the
  pre-push hook surfaced this: the suite runs *inside* a git hook, where the
  un-hermetic subprocess targeted the wrong repo).

## Tests

- Unit: ls-remote parsing + semver intent resolution; deterministic-compile
  reproducibility; the three lock flows (fake git IO); the `MSL-L215` gate.
- E2E (`tests/e2e/federated_dependency_test.ts`, fully offline, real `git`
  against a local bare repo): correct sha pin for `auto`/tag/branch, a second
  `lock` does not re-acquire, `check --strict` rejects a branch pin, and a
  cross-repo `Satisfies:` resolves — validated non-vacuous by a negative
  control (breaking the cache makes it fail with `MSL-T014` naming the ID).
- Full suite green: **3791 passed, 0 failed** (`just check`).

Each of the 10 implementation commits was TDD'd and independently reviewed; a
whole-branch review verified the four load-bearing invariants (warn-and-write
exit coherence, the producer↔consumer hash↔bytes contract, cross-machine
determinism, the `--strict` promotion) hold end-to-end.

## Deferred / accepted debt

- **Forge-tarball acquisition** (design §4.3 escalation) — filed as a
  fast-follow.
- **Non-blocking follow-ups** (filed as a debt issue): `lock --format json`
  omits ref/dep warnings from its structured output; a cosmetic cache-write
  message leaks the raw path; `makeTempDir` isn't wrapped as warn-and-write
  (a fatal env error); the idempotence e2e's sub-second mtime window.
- **Ungardened `docs/wip/`** — the epic design doc + this slice's plan remain
  in `docs/wip/` by design; gardening the epic into durable `docs/` records
  (ADR + guide) is slice 6's deliverable, matching slices 2 & 4.


stasson:
## `/review` — 8 finder angles + 1-vote verify (medium effort)

**No new Critical findings.** The two security issues surfaced during the
pre-merge whole-branch review — the git-URL transport-helper RCE and the
non-hermetic git subprocess (ambient `GIT_DIR` redirect) — were both fixed on
this branch before merge (`6ee73a2`, `b4bc2d1`). CI also caught a Windows-only
failure in the git-dependency e2e (custom `run()` helper resolved the CLI via
`URL.pathname` → malformed `/D:/…` path); fixed in `25ba938` by routing through
the shared `markspecInDir` helper.

Surviving findings, all Important-or-below, tracked in #785:

1. **[Important] Cross-kind upstream id collision** — `references:` and
   `dependencies:` share the `.markspec/cache/upstreams/<id>` namespace with no
   cross-kind dedup, so a shared derived `id` silently clobbers the cache and
   emits two rows under one id. Needs a collision diagnostic.
2. **[Consider] `lock` exits 0 on upstream resolution failure** — intended
   warn-and-write (decision 1), but per the repo's clig.dev convention an
   upstream warning arguably warrants exit 2; a declared-but-unreferenced
   dependency that fails to lock is otherwise invisible.
3. **[Minor]** `SHA_RE` rejects uppercase shas; hermetic `GIT_*` denylist is
   incomplete; `lock --format json` omits upstream warnings; `makeTempDir` not
   warn-and-write-wrapped.
4. **[Minor cleanup]** `l213`/`writeSnapshotCache` duplicate `upstream_refs.ts`;
   `l213(dir)` should be `l213(id)`; `.markspec.yaml` double-read in
   `acquire_compile`; unnecessary `remote add`; stringly-typed `resolved` pin
   kind.

The four load-bearing invariants (warn-and-write exit coherence, the
producer↔consumer hash↔bytes contract, cross-machine determinism, the
`--strict` promotion) were verified end-to-end and hold.

_Full detail and proposed fixes in #785._

Issue #745, opened 2026-07-04
story(lsp): LSP + MCP surfaces (federated upstream slice 5)


stasson:
Part of #741 (slice 5 of 6). Depends on slice 4 (graph integration +
validator).
Design: `docs/wip/2026-07-04-federated-upstream-resolution-design.md` §4.6,
§4.8.

## Goal

Surface upstream entries in the LSP and MCP the same way project entries
are surfaced, minus the write paths that don't apply to read-only citizens.

## Scope

- LSP completion: upstream display IDs offered with an
  `— from <upstreamId>@<version>` badge, reusing the existing `origin`
  plumbing (`formatEntryOrigin`) already wired for profile-delivered corpus.
- LSP hover / document highlights: work identically for upstream entries.
- LSP go-to-definition: a documented **no-op** for upstream entries — their
  `location.file` is an upstream-repo path that does not exist locally.
- Read-only decorations: rename stays blocked by the existing `origin`
  guard; `fmt`/`insert` cannot touch upstream entries (no local `.md` file
  backs them); diagnostics for them are never published — same pattern as
  the ADR-030 delivered-corpus guards, extended to the `upstream` origin
  kind.
- MCP: `entry_show` / `entry_list` / `entry_context` see upstream entries via
  the shared compile cache — no new MCP resource kind needed.

## Out of scope

- New MCP tools or resource kinds.
- Any write-path support for upstream entries — by design they are
  permanently read-only (design §4.6).

## Testing

- LSP unit/integration: completion item carries the origin badge; hover
  renders an upstream entry; document-highlight and rename correctly no-op
  or block on an upstream-origin entry; go-to-definition returns null/no-op
  without throwing.
- MCP: `entry_show`/`entry_context` on an upstream-origin display ID return
  the hydrated entry with its origin badge.

PR #787, opened 2026-07-05
fix(lsp): upstream entries have no navigable location — guard references, workspace symbols, code lens

### LINKED ISSUE #783: bug(lsp): upstream entries' relative location.file throws in pathToUri (onReferences, workspace/symbol)
## Summary

Locked-upstream entries (federated-upstream epic #741) carry a **tree-relative** `location.file` (e.g. `docs/product/stk.md`), because upstream projects are compiled with tree-relative paths for byte-reproducible snapshots (slice 3). Several LSP handlers map entries through `pathToUri(entry.location.file)` (i.e. `@std/path` `toFileUrl`), which **throws `TypeError: Path must be absolute`** on a relative path. Because those handlers iterate `index.getAllEntries()` — which includes seeded upstream entries since slice 4 — the request handler throws when an upstream entry is involved.

Confirmed:

```
$ deno eval 'import { toFileUrl } from "@std/path"; toFileUrl("docs/product/stk.md")'
TypeError: Path must be absolute: received "docs/product/stk.md"
```

`deserializeEntry` (`core/compiler/deserialize.ts`) reconstructs the upstream entry by spreading the serialized wire object verbatim, so `location.file` stays exactly as recorded in the upstream snapshot — relative.

## Affected surfaces (confirmed)

- **`onReferences`** (`lsp/server.ts`): `findReferencingEntries(index.getAllEntries(), id).map(entryToLspLocation)` plus the `includeDeclaration` branch `entryToLspLocation(decl)`. Find-All-References on an upstream display ID (VS Code sends `includeDeclaration: true` by default), or on any entry that an upstream entry references, throws.
- **`onWorkspaceSymbol`** (`lsp/symbols.ts` `entriesToWorkspaceSymbols`): builds `location.uri = pathToUri(entry.location.file)` for every matching entry over `index.getAllEntries()`. A workspace-symbol query that matches an upstream entry throws.

`onDefinition` was guarded in slice 5 (#745) via `resolveDefinitionLocation` (returns `null` for `isUpstreamEntry`). Document symbols are safe (no `uri`, and they use `getEntriesForFile` — local files only). Other `pathToUri(entry.location.file)` sites reachable via `getAllEntries()` (e.g. code lens / inlay hint dependent counting) should be audited.

## Severity

Pre-existing from slice 4 (#744) — when upstream entries first entered the LSP index. Not a regression from slice 5 (#745), which did not change `onReferences`/`onWorkspaceSymbol`. Not caught by tests because no LSP test seeds an upstream corpus and then exercises references/workspace-symbol. A thrown LSP request handler returns an error response to the editor (the server does not crash), so the failure mode is "Find-All-References / workspace symbol search silently errors when an upstream entry is involved."

## Proposed fix (pick one, applied at the right depth)

1. **Guard at the mapping sites** — route upstream entries through the same `resolveDefinitionLocation`-style null guard everywhere an entry is turned into an LSP `Location`, and drop upstream entries from `onReferences`/`onWorkspaceSymbol` results (consistent with the go-to-definition no-op: upstream entries have no local file to open). Consolidate the scattered guards rather than adding one-offs (relates to #771's "partition-entries-once" cleanup).
2. **Make `pathToUri` total** — a helper that returns `undefined` for a non-absolute/non-local path instead of throwing, so no call site can throw. Callers filter `undefined`.

Prefer a single shared decision point over per-handler special-cases.

## Links

- Epic #741 (federated upstream resolution), slice 4 #744 (seeded upstream into the LSP index), slice 5 #745 (guarded `onDefinition`).
- Discovered during the slice-5 PR #782 review.
- Related consolidation: #771.

stasson:
Closes #783.

Federated-upstream entries carry a **tree-relative** `location.file`; `pathToUri` (`@std/path` `toFileUrl`) throws `Path must be absolute` on it. Several LSP handlers mapped upstream entries — reachable via `index.getAllEntries()`, `findReferencingEntries`, a local entry's upstream `Satisfies:` target, or `index.getFilePaths()` — to `pathToUri(entry.location.file)` unguarded and **threw** when an upstream entry was involved (Find-All-References, workspace-symbol, and code lens on a cross-repo `Satisfies:`). Slice 5 (#745) guarded only `onDefinition`; this fixes the rest.

## Fix — one decision point + guard every exposed site

Only upstream entries carry relative paths, so the complete navigability predicate is `hasNavigableLocation(entry) = !isUpstreamEntry(entry)` (new, in `lsp/definition.ts`). Slice-5's `resolveDefinitionLocation` is renamed to `resolveNavigableLocation` (now used by both go-to-definition and find-references).

- **`onReferences`** — map referencing entries through `resolveNavigableLocation` and filter nulls; guard the `includeDeclaration` declaration too. Drops upstream referencers + an upstream declaration; all local references preserved.
- **`workspace/symbol`** (`entriesToWorkspaceSymbols`) — omit upstream entries (LSP `SymbolInformation` requires a location; an upstream entry has no local one).
- **`code_lens`** — an upstream `Satisfies:` target keeps its informative title but is non-clickable (`arguments: []`); upstream entries are filtered out of the "↑ dependents" reference locations. `code_lens.ts` stays `@std/path`-free by using the `isUpstreamEntry` semantic predicate directly.

`onDefinition` behavior is unchanged (pure rename — the old body already returned `null` for upstream). The now-unused `entryToLspLocation` import is dropped from `server.ts`.

## Verification

Linchpin test: a `pathToUri` stub that **throws** on a relative path, so any regression that lets a relative path through fails loudly; the `symbols` RED was reproduced through the real `@std/path` `toFileUrl`. `just build` green (3863 tests), fmt/dprint clean. An independent whole-file re-audit (opus) of every `pathToUri`/`location.file` site confirms **no unguarded upstream→pathToUri site remains** — each is guarded or provably operates on always-absolute local paths (open-document handlers only ever see local entries, since `getEntriesForFile` is queried with absolute paths and upstream entries are keyed under relative paths).

Accepted, documented nuance: the "↑ N dependents" count still counts upstream referrers while their locations are excluded from the click target (no-throw is the invariant that matters).

### DIFF
--- a/packages/markspec/lsp/code_lens.ts
+++ b/packages/markspec/lsp/code_lens.ts
@@ -24,6 +24,7 @@
  */
 
 import type { DisplayId, Entry } from "../core/mod.ts";
+import { isUpstreamEntry } from "../core/mod.ts";
 import { buildIncomingCount } from "./incoming_index.ts";
 import { findReferencingEntries } from "./references.ts";
 
@@ -84,11 +85,18 @@ export function buildCodeLenses(
     const depCount = incomingCount.get(entry.displayId) ?? 0;
     if (depCount > 0) {
       const referencing = findReferencingEntries(allEntries, entry.displayId);
-      const referenceLocations = referencing.map((ref) => ({
-        uri: pathToUri(ref.location.file),
-        line: Math.max(0, ref.location.line - 1),
-        character: 0,
-      }));
+      // Upstream entries (federated corpus, #783) carry a tree-relative
+      // location.file that pathToUri cannot convert — filter them out
+      // before conversion rather than throwing. depCount above still
+      // counts them; the rare count/location mismatch is acceptable, the
+      // invariant that matters is "no throw".
+      const referenceLocations = referencing
+        .filter((ref) => !isUpstreamEntry(ref))
+        .map((ref) => ({
+          uri: pathToUri(ref.location.file),
+          line: Math.max(0, ref.location.line - 1),
+          character: 0,
+        }));
       out.push({
         range,
         command: {
@@ -110,12 +118,18 @@ export function buildCodeLenses(
         if (target) {
           const targetLine = Math.max(0, target.location.line - 1);
           const targetPos = { line: targetLine, character: 0 };
+          // A resolved-but-upstream target keeps its informative title but
+          // becomes non-clickable — its location.file is a tree-relative
+          // path pathToUri cannot convert (#783).
+          const navigable = !isUpstreamEntry(target);
           out.push({
             range,
             command: {
               title: `↓ Satisfies: ${tok} — ${target.title}`,
               command: "markspec.openDefinition",
-              arguments: [pathToUri(target.location.file), targetPos],
+              arguments: navigable
+                ? [pathToUri(target.location.file), targetPos]
+                : [],
             },
           });
         } else {
--- a/packages/markspec/lsp/code_lens_test.ts
+++ b/packages/markspec/lsp/code_lens_test.ts
@@ -208,6 +208,98 @@ Deno.test("buildCodeLenses: entry with multiple Satisfies values yields one lens
   assertEquals(satLenses[1].command?.title, "↓ Satisfies: STK_B — B");
 });
 
+// --- upstream entries (#783) ---
+//
+// Federated-upstream entries carry a tree-relative location.file (e.g.
+// "docs/x.md") — pathToUri (@std/path toFileUrl, or any well-behaved
+// implementation) cannot convert a relative path. This stub THROWS on a
+// relative path, so any of these tests would fail if an unguarded relative
+// path reached it — the linchpin proving no relative path reaches
+// pathToUri at all.
+const throwingPathToUri = (p: string): string => {
+  if (!p.startsWith("/")) {
+    throw new Error(`relative path reached pathToUri: ${p}`);
+  }
+  return "file://" + p;
+};
+
+/** Build an upstream-origin fixture on top of {@linkcode fakeEntry}. */
+function fakeUpstreamEntry(opts: {
+  displayId: string;
+  title: string;
+  file: string;
+  line: number;
+  satisfies?: string;
+}): Entry {
+  return {
+    ...fakeEntry(opts),
+    origin: { kind: "upstream", upstreamId: "product", version: "v1" },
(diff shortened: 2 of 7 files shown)


stasson:
## Review — independent completeness re-audit (this bug class had prior reviews under-rate it)

An independent whole-file re-audit (opus) of **every** `pathToUri` / `entry.location.file` → URI site in `lsp/`, classifying each as guarded / provably-safe / exposed:

| Site | Can be upstream? | Verdict |
|---|---|---|
| `definition.ts` `entryToLspLocation` / `resolveNavigableLocation` | yes | guarded (`hasNavigableLocation`) |
| `code_lens.ts` top-of-loop `uri` | no (open-doc entries, absolute keys) | provably-safe |
| `code_lens.ts` dependents `referenceLocations` | yes (`findReferencingEntries(allEntries)`) | guarded (`!isUpstreamEntry` filter) |
| `code_lens.ts` Satisfies `target` | yes (`byDisplayId` from `allEntries`) | guarded (title kept, `arguments: []`) |
| `symbols.ts` `entriesToWorkspaceSymbols` | yes (`getAllEntries()`) | guarded (`hasNavigableLocation` skip) |
| `server.ts` `publishAllDiagnostics` / `onRenameRequest` (`getFilePaths()`) | yes | guarded (pre-existing `corpusFilePaths`; `seedUpstreamCorpus` registers upstream paths there) |
| `server.ts` `publishFileDiagnostics` / `onDocumentLinks`; `symbols` document-symbols; `inlay_hint` / `document_links` / `incoming_index` / `entry_ranges` | no (open-doc / absolute / no URI built) | provably-safe |

**Result: no still-exposed site remains.** The two families the earlier review under-rated — anything fed `index.getAllEntries()` and anything iterating `index.getFilePaths()` — are all covered. Open-document handlers are provably safe because `getEntriesForFile` is only ever queried with an absolute path and upstream entries are keyed under relative paths, so they can never appear in an open document's entry list.

Linchpin test verified: `code_lens_test.ts` injects a `pathToUri` stub that throws on any relative path, so a regression that lets one through fails loudly; the `symbols` RED was reproduced through the real `@std/path` `toFileUrl`. `just build` green (3863 tests).

**Verdict: Ready to merge.** 0 Critical / 0 Important. One documented, accepted Minor: the "↑ N dependents" count still counts upstream referrers while their click-locations are excluded (no-throw is the invariant).

PR #779, opened 2026-07-04
feat(core): uxil parser — ux: reference grammar + declaration forms (#725)

stasson:
Closes #725.

Part of the uxil epic (#717 §B). First uxil slice — **parse only**.

## What this adds

A new `packages/markspec/core/uxil/` layer that turns uxil declaration and
`ux:` reference source strings into a typed declaration AST plus structured
parse diagnostics. It mirrors the sibling typl DSL's *parse* half
(`lexer → grammar → ast + diagnostics`) and rides the same intent, stopping
exactly where typl's `assemble.ts` / `validator.ts` begin — those map to
S8/S9.

| File | Responsibility |
| --- | --- |
| `ast.ts` | `UxRef`, `UxKey`, `RootDecl`, `ElementDecl`, `ChildSurfaceDecl`, `UxilDecl`, `Position` |
| `diagnostics.ts` | `UXIL_CODES` (UXIL-001…008) + `uxilDiagnostic()` |
| `lexer.ts` | single-line `tokenize()` |
| `grammar.ts` | `parseUxRef`, `parseRootDecl`, `parseChildSurfaceDecl`, `parseElementBullet` |
| `recognize.ts` | `classifyUxilForm()` — routes a code span to a form |
| `mod.ts` | barrel |

## The grammar

Reference: `["ux:"] surface ["@" state] ["/" element [":" key] ["!" verb]]`, with
`surface = segment *("." segment)`. The **scheme-less wire form**
(`media.home/play`) parses byte-identically to `ux:media.home/play` except for
`hasScheme` — the wire-compat hard constraint, covered by a test.

Four declaration forms parse and round-trip: root inline
(`ux:media.home : screen @ loading, error, ready`), element bullet
(`` `/play : activate` — event dictionary``), child surface
(`.confirm_dialog @ default`), and refs. `?`/`#` and the `ux://` authority
form are reserved (UXIL-002 / UXIL-003).

## Scope boundary (deferred to S8/S9)

- No entry-body walking, base resolution, single-root/duplicate checks,
  kind-inheritance, or uxRegistry — all **S8 (#726)**.
- No diagnostic bridging into `check`/`lint`/LSP — **S9 (#727)**.
- `core/uxil/` is intentionally **not** exported from `core/mod.ts` yet:
  built-and-tested but unreferenced, same as typl's parse pieces before their
  assemble landed.

## ⚠️ One decision to sanity-check — grammar decision K1

The issue is silent on where an element's **key template** sits, because `:`
[...]

Issue #730, opened 2026-07-03
story(docs): uxil ADR + ADR-019 namespacing update + guide chapter


stasson:
Part of #717. Depends on: S5, S8 (document what shipped).

## Goal
- New **ADR** for uxil (sibling of ADR-019): reference grammar, closed vocabularies + extension criteria, base-resolution, alternatives.
- Update **ADR-019** for the typl published/namespaced tier (§C.1) + TYPL-002/003 retirement.
- Guide chapter: authoring uxil declarations, `ux:` refs, typl namespacing, payload bridge.

## Acceptance
- ADRs merged; ADR-019 "Implementation status" reflects the namespacing tier.
- Guide chapter with worked examples; linked from the epic + design record.

Issue #725, opened 2026-07-03
story(uxil): parser — ux: URI reference grammar + declaration forms


stasson:
Part of #717 §B. Depends on: S2 (shared machinery), S4 (base-resolution). First uxil slice — parse only.

## Goal
Parse uxil declarations + citations on the shared machinery.

**Reference grammar** — every ref is a valid RFC 3986 URI, scheme `ux`:
```
ux-ref  = "ux:" surface ["@" state] ["/" element [":" key] ["!" verb]]
surface = segment *("." segment)          ; dots = containment
segment = lower *(lower / digit / "_")
key     = concrete-value / "{" name "}"
```
`?`/`#` reserved; `ux://app/…` authority reserved for a future owning-app qualifier. The scheme-less form is a valid relative URI reference, byte-identical to the deployed wire format (`media.home/play`) — **wire compatibility is a hard constraint**.

**Declaration forms** (inside the profile-designated contract entry type):
- Root inline span, exactly one: `` `ux:media.home : screen @ loading, error, ready` ``
- Element bullet: `` `/play : activate` `` (leading slash; verb set; optional key template; `@state`; `-> nav target`; mandatory trailing prose = event dictionary)
- Child surface bullet: `` `.confirm_dialog @ default` `` (leading dot; nested bullets = its elements; kind inherited)
- Table rows ≡ bullets; caption base per §A.

## Acceptance
- Grammar parses/round-trips all four forms + refs to the declaration AST.
- Scheme-less relative form parses identically (wire-compat test).
- Parse errors structured (feed S9).

PR #790, opened 2026-07-05
fix(core): uxil element key template becomes a clause after the verb set

### LINKED ISSUE #786: uxil parser: confirm element key-template placement (K1 — `/track{id}` vs `/track:{id}`)
Follow-up from S7 (#725, PR #779). Grammar decision **K1** was made during
implementation and needs confirming against the epic's design doc (in the seed
repo, not visible from this repo).

## The ambiguity

The `:` character means different things in a reference vs. a declaration:

- **Reference** (`ux-ref` grammar): `:` after an element introduces a **key** —
  `ux:media.home/track:{id}`.
- **Element declaration** (bullet): `:` introduces the **verb set** —
  `` `/play : activate` ``.

Issue #725 says an element declaration may *also* carry an optional **key
template**, but does not specify where. Reusing the reference's `element:{key}`
form inside a declaration would put two `:` in conflict:

```
`/track:{id} : activate`
        ^        ^
        key?     verbs?     ← ambiguous
```

## What S7 shipped (the decision to confirm)

The key template is attached in **braces directly after the element name, with
no colon**, so the single `:` is unambiguously the verb set:

```
`/track{id} : activate`   ->  element="track", keyTemplate={template:"id"}, verbs=["activate"]
```

Implemented in `parseElementBullet` (`packages/markspec/core/uxil/grammar.ts`).

## Action

Check this against the epic design doc:

- If the doc specifies **`/track{id}`** (braces-after-name) — no change, close
  this issue, and record the rule in the eventual uxil spec/guide.
- If the doc specifies **`/track:{id}`** (ref-style colon) — change it. The fix
  is localized: the key-template branch in `parseElementBullet` plus its unit
  test; no impact on the AST shape or the other parsers.

Part of #717 (uxil epic).

stasson:
Closes #786.

## What

Resolves grammar decision **K1** from S7 (#725 / PR #779) by conforming
`parseElementBullet` to the epic's design doc. The element
declaration's optional key template moves from braces glued to the
element name to its own `:` clause after the verb set:

```text
before (shipped):  `/track{id} : activate`
after (this PR):   `/track : activate : {id}`
doc's example:     `/favorite_toggle : toggle : {track_id}`
```

## Why

#786 anticipated the doc saying either `/track{id}` (shipped) or
`/track:{id}` (ref-style). The doc's worked example actually uses a
third form — the key template as a spaced `:` clause *after* the verb
set. Confirmed with the maintainer; conforming to the doc was chosen
because:

- it keeps one mental model across all declaration surfaces (identity
  first, then spaced property clauses — same as root decls and typl's
  `$name : kind`);
- the shipped glued form created a near-miss spelling with the citation
  form (`track_row{track_id}` vs `track_row:{track_id}`);
- the K1 ambiguity never arises in the doc's grammar: the first `:`
  clause is always the verb set, and a `{…}`-shaped clause is always
  the key.

## How

Localized exactly as #786 predicted: the key-template branch in
`parseElementBullet` moves from after-the-element to a clause slot
after the verbs. The old glued form now fails with `UXIL-005`. No AST
change (`ElementDecl.keyTemplate` unchanged); `parseUxRef`,
`parseRootDecl`, and `parseChildSurfaceDecl` untouched — the ref
grammar keeps `element:{key}`.

Clause order when combined: `/element : verbs [: {key}] [@ states]
[-> nav]` (key immediately after verbs, mirroring the doc's clause
order). Recording the rule in the uxil spec/guide remains a doc task
for a later story (S12 #730).

## Tests

TDD: rewrote the combined element-bullet test to the clause form, added
the doc's literal `/favorite_toggle : toggle : {track_id}` example, a
rejection test for the old glued form, and a missing-key edge
(`: activate :` → UXIL-001) — watched the three form-changing tests
fail red before the fix. Full `just build` green (lint + tests +
typecheck + compile); pre-push `just check` green.

### DIFF
--- a/docs/wip/2026-07-05-uxil-parser-s7-design.md
+++ b/docs/wip/2026-07-05-uxil-parser-s7-design.md
@@ -95,8 +95,9 @@ marks a root **declaration**. A `:` that appears _after_ `/element` is a ref
 ```
 
 - Leading `/` → element name (`play`).
-- Optional key template (`{id}`) on the element.
 - `: activate[, …]` → **verb set** (one or more verbs; comma list).
+- Optional key template as its own `: {key}` clause after the verb set
+  (`/track : activate : {id}` — resolved by #786, PR #790).
 - Optional `@state[, …]` → state set.
 - Optional `-> <ux-ref>` → **nav target**, parsed as a `UxRef` (may be
   scheme-less / relative).
--- a/docs/wip/2026-07-05-uxil-parser-s7-plan.md
+++ b/docs/wip/2026-07-05-uxil-parser-s7-plan.md
@@ -951,6 +951,12 @@ activate`), NOT after a colon — the colon is reserved for the
 verb set, so the ref-grammar's `/element:{key}` form cannot be reused here
 without a double-colon clash.
 
+> **Resolution (#786, PR #790):** checked against the seed design doc, which
+> uses a third form — the key template is its own `:` clause **after the verb
+> set** (`/favorite_toggle : toggle : {track_id}`). The braces-after-name form
+> described above was reversed; the shipped shape is
+> `/element : verb[, verb…] [: {key}] [@state, …] [-> nav-ref]`.
+
 - [ ] **Step 1: Write the failing tests (append to grammar_test.ts)**
 
 ```ts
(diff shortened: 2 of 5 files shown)


stasson:
@/tmp/c790_new.md

PR #791, opened 2026-07-05
fix(core): resolve federated upstream slice-3 review follow-ups (#785)

### LINKED ISSUE #785: Federated upstream slice 3 — review follow-ups & deferred items (#743)
Follow-ups from the slice-3 (git dependency fetcher) work and the `/review`
of #784 (merged as slice 3 of epic #741). None are merge-blockers — the two
security bugs the review surfaced (git-URL transport-helper RCE, non-hermetic
git subprocess) were fixed before #784 merged. These are the non-blocking
remainder, most-actionable first.

## Important — cross-kind upstream id collision (silent cache clobber)

`resolveProjectReferences` and `resolveProjectDependencies` each dedup ids
*within* their own kind, but both write into the same
`upstreamCacheRoot(projectRoot)/<id>` namespace with **no cross-kind check**.
A `references:` entry and a `dependencies:` entry whose URLs derive the same
`id` (via `deriveUpstreamId`) will:

- clobber each other's `manifest.json`/`compiled.json` on disk (whichever
  resolver runs second wins), and
- write both a `[[upstream.registry]]` and a `[[upstream.dependency]]` row
  with the same `id`, which `upstreamRefsFromLockfile` then maps to two
  `UpstreamSnapshotRef`s pointing at the same `dir` → duplicate/nondeterministic
  corpus entries.

No diagnostic fires. **Proposed fix:** detect an id used by both a resolved
reference and a resolved dependency and emit an `MSL-R014`-style collision
diagnostic (warn-and-write: skip the colliding dependency row), e.g. pass the
resolved reference ids into `resolveProjectDependencies` to dedup across kinds.
Add a references+dependencies id-collision e2e case.

## Consider — `markspec lock` exit code on upstream warnings

With warn-and-write (decision 1), a `references:`/`dependencies:` failure is a
warning and `lock` writes the resolvable pins, but the final exit is keyed only
on `resolved.diagnostics` — so `lock` exits **0** even when an upstream failed
to resolve and its row was dropped. `lock --check` (the CI drift gate) is
unaffected, and a broken cross-repo ref still surfaces at `check` time — but a
*declared-but-unreferenced* dependency that silently failed to lock is invisible.
Per the repo's clig.dev convention ("2 for warnings only"), `lock` emitting
upstream warnings should arguably exit **2** (still writing the lockfile) so CI
retains a signal. Needs a small policy decision since it changes the exit
contract.

## Minor — correctness/robustness polish

- `git_intent.ts` `SHA_RE = /^[0-9a-f]{40}$/` rejects a valid uppercase/mixed-case
  40-hex sha passed as a `version:` intent → relax to `/^[0-9a-fA-F]{40}$/`.
- Hermetic git env in `lock.ts` uses a 7-var `GIT_*` **denylist**
  (`runGit`) — inherently incomplete (`GIT_CEILING_DIRECTORIES`,
  `GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_CONFIG`, … uncovered). Either extend
  the denylist or switch to an allowlist of the few vars git actually needs.
- `lock.ts` `makeTempDir` isn't wrapped as warn-and-write — a temp-dir creation
  failure aborts the whole lock (accepted as a fatal env error, but a defensive
  wrap would tighten the invariant).
- `lock --format json` omits `refResult`/`depResult` warnings from its structured
  output (only `resolved.diagnostics` is serialized) — agents consuming JSON
  don't see that an upstream failed.

## Minor — cleanup

- `upstream_deps.ts` `l213()` duplicates `upstream_refs.ts`'s `l213()`; and
  `writeSnapshotCache()` duplicates `writeCache()` — extract shared helpers
  (parameterize the "reference"/"dependency" noun).
- `writeSnapshotCache` passes the full cache `dir` (not the clean `id`) to
  `l213`, so a cache-write-failure message leaks the raw path.
- `resolveProjectDependencies` repeats `if (existing) dependencies.push(existing);
  continue;` after each error branch — collapse to a single push/continue.
- `acquire_compile.ts` reads+parses `.markspec.yaml` twice
  (`loadProfileForCommand` + `loadToolConfig`) per acquired dependency.
- `denoGitIO.acquireTree` uses `init`+`remote add`+`fetch` — `git fetch <url>
  <sha>` works directly, dropping one subprocess.
- `runGit` rebuilds the curated env per call (~5×/dependency) — compute once.
- `resolved` pin kind is stringly-typed (`"tag:x"`, matched by
  `startsWith("tag:")`) while the sibling `kind` field uses a discriminated
  union — a `{ kind: "tag"|"branch"|"sha"; name }` shape would make
  `dependencyPinAssurance` exhaustive/refactor-safe (touches the shipped
  `UpstreamDependency` model, so scope carefully).
- Determinism of the compiled snapshot rests on the *absence* of stat/git
  callbacks + one regression test rather than a structural guarantee.

## Deferred feature — forge-tarball acquisition (design §4.3)

v1 acquires via shallow `git fetch`-by-sha (decision 2). The design's
owner-publishes escalation — prefer the GitHub/GitLab tarball endpoint for a
recognized host (one authed API request, no `.git`) — is deferred; it needs an
authed host and can't be exercised by the offline e2e.

stasson:
Closes #785.

Resolves the non-blocking follow-ups from the `/review` of #784 (federated
upstream slice 3, the git dependency fetcher). The two security criticals were
fixed before #784 merged; this PR clears the remaining Important + Minor items,
defers the two genuinely-deferred ones, and updates the catalogue + CLI docs.

## Important

- **Cross-kind id collision (`MSL-L216`).** A `dependencies:` entry deriving an
  id already claimed by a `references:` entry no longer silently clobbers the
  shared `.markspec/cache/upstreams/<id>` cache namespace or double-maps the
  corpus. Declared reference ids flow into `resolveProjectDependencies` via a
  `reservedIds` set; the colliding dependency is skipped with a warn-and-write
  `MSL-L216` warning.

## Behavior

- **`markspec lock` exit code.** Now exits **2** (still writing the lockfile)
  when an upstream *acquisition* warning fires (`MSL-L213`/`L214`/`L216`), per
  the repo's clig.dev "2 for warnings only" rule, so a bare `markspec lock` in
  CI surfaces a dropped or stale pin. Benign resolution advisories (`MSL-L102`
  identity-hash for an unresolvable bundled profile) deliberately do **not**
  gate the exit code. `lock --format json` now includes the reference +
  dependency warnings, not just the resolved-set diagnostics.
- **Uppercase sha.** `git_intent` accepts a mixed-case 40-hex `version:` sha
  and normalizes the pin to lowercase — determinism: no lockfile differs on sha
  case alone.

## Robustness / cleanup

- Hermetic git env extended from 7 repo-discovery vars to also scrub
  object-store (`GIT_ALTERNATE_OBJECT_DIRECTORIES`, `GIT_CEILING_DIRECTORIES`)
  and config-injection (`GIT_CONFIG*`) redirects; the curated env is computed
  once per process rather than per git call.
- `acquireTree` fetches by URL directly — drops the redundant `remote add` and
  the pointless `--filter=blob:none`. `compileAcquiredTree` reads the whole
  tree, so `checkout` refetched every filtered blob anyway; the filter saved
  nothing, and its promisor remote was the only reason `remote add` existed
  (keeping `blob:none` while dropping `remote add`, as the issue literally
  suggested, would have broken checkout).
- A `makeTempDir` failure is now warn-and-write rather than aborting the lock.
- Shared `l213` / cache-writer extracted to `upstream_common.ts` (fixes the
  `dir`→`id` path leak in the dependency cache-write warning); collapsed the
  repeated push-existing/continue branches via `restore()`/`firstLockOrUpdate()`
  helpers; memoized the config/profile reads in `compileAcquiredTree` (kills the
  double `.markspec.yaml` read per acquired dependency).

## Deferred (kept on #785)

- The discriminated-union refactor of the `resolved` pin field — the field is
  shared across registry/profile/dependency rows (plain version strings vs
  `tag:`/`branch:`/`sha:`-prefixed), so converting one side diverges the wire
  format for a one-line `startsWith("tag:")`. Low value for the blast radius.
- Forge-tarball acquisition (design §4.3) — needs an authed host, can't be
  exercised by the offline e2e.

## Tests

- **Unit:** cross-kind collision (`MSL-L216`, dependency skipped, cache not
  clobbered), non-colliding `reservedIds` no-op, uppercase-sha lowercased.
- **E2e:** cross-kind collision (registry kept, dependency dropped, exit 2),
  unreachable-dependency exit-2 warn-and-write; the restore-mismatch scenario
  updated to the new exit-2 contract.
- `just build` + both format gates green locally; pre-push `just check` green
  (3862 passed, 0 failed).

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -7,7 +7,8 @@ Every command supports `--help`. Commands that produce structured output support
 `--format json` for machine-readable output to stdout (diagnostics always go to
 stderr).
 
-**Exit codes:** `0` success, `1` error, `2` warnings only (`check`, `lint`).
+**Exit codes:** `0` success, `1` error, `2` warnings only (`check`, `lint`,
+`lock`).
 
 **Global options** (available on every command):
 
@@ -771,6 +772,15 @@ done by hand.
 | `MSL-L214` | `markspec lock` (restore flow) | The cache needed restoring, but the re-fetched content's hash no longer matches the pinned snapshot — the published site moved. Run `markspec lock --update=<id>` to move the pin deliberately.                                                                                             |
 | `MSL-L212` | `markspec check` (offline)     | Also covers upstream cache drift in this release: fires when a locked reference's cache under `.markspec/cache/upstreams/<id>/` is missing or its hash no longer matches `markspec.lock`, in addition to the existing traceability-edge-drift case. Either way, the fix is `markspec lock`. |
 | `MSL-L215` | `markspec check`               | A `dependencies:` pin resolved to a branch or bare sha rather than a tag — an unreleased state. Advisory by default; `markspec check --strict` promotes it to a hard error, so a release build cannot pass against an unbaselined dependency.                                               |
+| `MSL-L216` | `markspec lock`                | A `dependencies:` entry derives the same upstream id as a `references:` entry. The dependency is skipped — the reference snapshot owns the shared `.markspec/cache/upstreams/<id>/` namespace — so set a distinct `name:` on one of them. Warn-and-write: every other pin still locks.      |
+
+**Exit codes.** `markspec lock` exits `0` when every upstream resolved cleanly.
+Following the clig.dev convention (`2` for warnings only), it exits `2` — while
+still writing `markspec.lock` (warn-and-write) — when any upstream could not be
+cleanly locked (`MSL-L213`/`MSL-L214`/`MSL-L216`), so a bare `markspec lock` in
+CI surfaces a dropped or stale pin. A hard error (e.g. an invalid `project.yaml`
+or sync mapping) exits `1`. `markspec lock --check` is unaffected: it stays a
+read-only drift gate that exits `1` on drift.
 
 **`dependencies:` are acquired and compiled during `lock`.** For each declared
 `dependencies:` entry, `markspec lock` resolves the declared version intent
--- a/docs/spec/language/language.md
+++ b/docs/spec/language/language.md
@@ -1647,12 +1647,16 @@ The lockfile (`MSL-L###`) and external-sync (`MSL-S###`) diagnostic families are
 governed by [ADR-022](../../architecture/adr-022-lockfile-and-external-sync.md)
 and catalogued in
 [ADR-012](../../architecture/adr-012-diagnostic-code-scheme.md) (ADR-022
-amendment). The offline composite `markspec check` gate emits one of them:
+amendment). The offline composite `markspec check` gate and `markspec lock` emit
+these lockfile diagnostics:
 
 | ID         | Severity | Rule                                                                                                                                                                                                                                                               |
 | ---------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `MSL-L212` | error    | Canonical edge-hash drift: the project's traceability edges no longer match the hash pinned in `markspec.lock` (run `markspec lock`). Fires only when `markspec.lock` exists; checked offline (no network — upstream resolution stays in `markspec lock --check`). |
+| `MSL-L213` | warning  | An upstream `references:` or `dependencies:` entry could not be locked (unreachable remote, malformed manifest, cache-write failure). Warn-and-write: the resolvable pins still lock. Emitted by `markspec lock`, which then exits 2.                              |
+| `MSL-L214` | warning  | A `references:` restore fetched a snapshot that no longer matches the locked pin (the published site moved). The pin is kept unchanged — run `markspec lock --update=<id>` to move it. Emitted by `markspec lock`, which then exits 2.                             |
 | `MSL-L215` | warning  | A `dependencies:` pin resolved to a branch or bare sha, not a tag (an unreleased state). Advisory by default; promoted to an error under `markspec check --strict` — release builds require every dependency to be tag-pinned.                                     |
+| `MSL-L216` | warning  | An upstream id is claimed by both a `references:` and a `dependencies:` entry. The dependency is skipped (the reference snapshot owns the shared cache); set a distinct `name:` on one of them. Emitted by `markspec lock`, which then exits 2.                    |
 
 > **Prefix overlap (known, pre-existing).** The `MSL-L###` lockfile family
 > shares its prefix with §8.4's link rules (`MSL-L006`), and the `MSL-S###`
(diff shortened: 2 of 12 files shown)


stasson:
## `/review` — 8 finder angles + analysis

**Zero Critical, zero merge-blocking findings.** The correctness angles
(line-by-line, removed-behavior, cross-file) confirmed the load-bearing parts
are sound:

- The collision-skip ordering and the exit-2 logic — `MSL-L213`/`L214`/`L216`
  correctly land in `refResult`/`depResult` and gate exit 2, and a branch-pinned
  dependency does **not** spuriously trip it (`MSL-L215` is emitted only at
  `check` time, not during `lock`).
- The `acquireTree` fetch simplification — identical `allowReachableSHA1InWant`
  requirement, improved non-partial-clone-server compatibility, and (verified)
  the whole tree is checked out so `--filter=blob:none` saved nothing while its
  promisor remote was the only reason `remote add` existed.
- The shared `upstream_common.ts` extraction (byte-identical reference wording;
  the deliberate dependency `dir`→`id` message fix), the `curatedGitEnv` +
  `compileAcquiredTree` memoizations, and the `git_intent` sha-lowercasing (all
  consumers case-agnostic).

Reuse + conventions angles clean (no `Deno.*` in core, `core/mod.ts` boundary
respected, doc comments present, docs tables conform).

The surviving observations are design/hardening follow-ups and pre-existing
gaps, tracked in #793:

1. **[hardening]** Env scrub keeps direct command-exec vars
   (`GIT_SSH_COMMAND`, `GIT_ASKPASS`, …) — a deliberate asymmetry (they are
   auth-bearing; a hostile parent already has code-exec) worth documenting + an
   allowlist evaluation.
2. **[design, pre-existing]** Dependency intent-drift is silent: the KEEP branch
   and `checkDrift` don't warn when `version:` is edited but the cache still
   validates.
3. **[altitude]** Exit-code gate keys on the source bucket rather than the
   diagnostic's own category.
4. **[design]** A `makeTempDir` failure (systemic) is warn-and-write → pin-less
   lockfile + exit 2; arguably should be fatal.
5. **[altitude]** Cross-kind collision uses a one-directional reserved-set;
   a single pass over all upstream kinds would be more robust.
6. **[minor]** ADR-012 catalogue omits L213–L216 (pre-existing; catalogued in
   `language.md` §8.10 instead); sequential upstream resolution (pre-existing
   perf); `intent` column not case-normalized (cosmetic).

Plus the two items #785 deferred by design (discriminated-union `resolved`
refactor; forge-tarball acquisition), also on #793.

_Merging: CI green on macOS / ubuntu / windows-latest, no Critical findings._

Issue #727, opened 2026-07-03
story(uxil): UXIL-0xx diagnostics family


stasson:
Part of #717 §B. Depends on: S8 (compiler) — co-develops. Sibling of TYPL-0xx.

## Goal
Define the `UXIL-0xx` diagnostics family:
unknown surface/element; verb not in declared set; kind violation (navigate to / visibility of a non-screen; states on a stateless kind); duplicate surface declaration; dangling namespace parent; undeclared state referenced; relative ref without base; zero/multiple root declarations; concrete key where a template is required; declaration outside the declaring entry type.

## Acceptance
- Each code has a template, severity, and a triggering test fixture.
- Registered in the catalogue (ADR-012 scheme) with LSP `codeDescription` targets reserved.

PR #796, opened 2026-07-05
fix(core): uxil parser emits one specific diagnostic per malformed input

### LINKED ISSUE #780: uxil parser: redundant / contradictory parse diagnostics on malformed input
Follow-up from the #725 / #779 code review (S7 uxil parser). Several malformed
inputs produce **redundant or contradictory** parse diagnostics. Not
correctness-critical — the parser still flags the error — but noisy, and S9
(#727) will surface these to users in the editor, so they should be cleaned up
before or with the diagnostics wiring.

Root cause for most: the parser continues after pushing a specific diagnostic,
and a later `expectEof` (or a second scan) re-flags the same defect.

Cases (all in `packages/markspec/core/uxil/grammar.ts`):

1. **Empty verb set → two codes.** `` `/play :` `` emits UXIL-001 "expected verb"
   (from `expectIdent`) **and** UXIL-005 for the same empty verb set.
2. **Reserved char in the `-> nav` tail → double UXIL-002.** `scanReservedChars`
   scans the whole span and `parseUxRef(navSource)` scans the nav again, so a
   single `?`/`#` in the nav target is reported twice.
3. **Malformed surface → UXIL-008 + spurious UXIL-001.** `parseUxRef("media..home")`
   yields UXIL-008 then a spurious "unexpected 'home'". Same for a reserved char
   mid-ref (`media.home?x` → UXIL-002 + spurious UXIL-001, because the lexer
   drops the char and orphans the trailing token).
4. **Missing colon but verb present → contradictory UXIL-005.**
   `` `/play activate` `` reports UXIL-005 "empty verb set" even though the verb
   is present; the real defect is the omitted `:`.
5. **Missing kind → inconsistent code + return shape.**
   `parseRootDecl("ux:media.home")` → UXIL-004 + no decl, but
   `parseRootDecl("ux:media.home :")` → UXIL-001 + a decl with `kind=""`.
6. **Surface literally named `ux`.** `consumeScheme` greedily treats a leading
   `ux:` as the scheme, so `parseRootDecl("ux : screen")` misparses (spurious
   UXIL-004, no decl). Documented as a known limitation in the S7 design doc;
   revisit if a real `ux`-named surface is ever needed.

Suggested direction: after emitting a specific diagnostic on a hard parse
error, stop the parser rather than letting `expectEof` re-flag; scan reserved
chars once (not again inside a nested `parseUxRef`); and make the
missing-kind / empty-verb / missing-colon cases each emit exactly one, correct
code with a consistent return shape.

Part of #717 (uxil epic). Blocks clean UX for #727 (S9 diagnostics).

stasson:
Closes #780.

Follow-up from the #725/#779 (S7 uxil parser) review: malformed inputs produced redundant or contradictory parse diagnostics. Root cause: the parser kept going after pushing a specific diagnostic, and a later `expectEof` (or a second reserved-char scan) re-flagged the same defect. S9 (#727) will surface these codes in the editor, so each malformed region now reports exactly one, correct code.

## Changes (all in `core/uxil/grammar.ts`)

| # | Case | Before | After |
| - | ---- | ------ | ----- |
| 1 | `` `/play :` `` empty verb set | UXIL-001 + UXIL-005 | UXIL-005 |
| 2 | reserved char in `-> nav` tail | UXIL-002 ×2 | UXIL-002 ×1, span-relative column |
| 3 | `media..home` / `media.home?x` | UXIL-008 / UXIL-002 + spurious UXIL-001 | the specific code only |
| 4 | `` `/play activate` `` (missing `:`, verb present) | contradictory UXIL-005 "empty verb set" | UXIL-001 `expected ':' before the verb set` |
| 5 | `ux:media.home` vs `ux:media.home :` (missing kind) | UXIL-004 + no decl vs UXIL-001 + decl `kind:""` | both: UXIL-004 + partial decl (`kind:""`, surface + states captured) |
| 6 | surface literally named `ux` | known limitation | unchanged (breadcrumb added on `consumeScheme`) |

Post-review hardening (see the /review comment below; found by the review, fixed in the amended commit):

- A reserved `?`/`#` now **stops structural parsing outright** in all four parsers — the lexer drops the char, so structure checks were firing on a corrupt token stream (e.g. `[UXIL-002, UXIL-004]` for `media.home?x` as a root decl).
- The peeled `-> nav` tail is scanned on **every** return path (a struct-part failure no longer swallows a reserved char in the nav).
- Nav diagnostics are re-based to **span-relative columns** at the splice point, so S9 needs no nav-specific offset handling.
- The load-bearing `expectEof`-before-nav/dictionary ordering is documented at the call site and pinned by a test.

## Design decisions (ratified with @t before implementation)

- **Suppression rule:** `expectEof` no-ops once any diagnostic exists — one clear structural error per region. Editors reparse per keystroke, so errors surface progressively. Orthogonal defects in different regions (e.g. UXIL-001 trailing junk + UXIL-006 missing event dictionary) still co-report.
- **Case 4 → UXIL-001 + detail** rather than a new code: keeps the taxonomy (001 = structural, 005 = genuinely empty verb set) and the catalogue at 001–008; a dedicated code can be minted when S9 builds an "insert `:`" quick fix.
- **Case 5 → best-effort partial decl:** honors the documented "decl undefined only when no surface" contract and matches `parseElementBullet`'s pattern (decl with `verbs: []` + UXIL-005). Mid-edit, editor tooling keeps the surface/states while the squiggle points at the missing kind. **S8 contract note (#726):** assemble must gate uxRegistry registration on a clean parse — recorded in `parseRootDecl`'s doc comment and on #726.

## Notes for reviewers

- Composes with the #786 key-clause rework: all its exact-code-list tests still pass unchanged.
- Reserved-char inputs no longer return a best-effort decl/ref (the stream is corrupt); reserved chars are rare by construction, so the trade favors diagnostic honesty.

## Testing

- 12 focused tests (one per issue case + one per review finding; the pre-existing `.some()` assertions tightened to exact code lists), TDD red→green in both rounds. Module suite: 36 passed.
- `just build` (lint + test + typecheck + compile), `deno fmt --check`, and `dprint check` all pass.

### DIFF
--- a/packages/markspec/core/uxil/grammar.ts
+++ b/packages/markspec/core/uxil/grammar.ts
@@ -5,6 +5,16 @@
  * ({@linkcode parseUxRef}) plus the three declaration-form parsers (added in
  * later tasks). Each returns a best-effort AST node and a list of
  * source-local {@linkcode UxilDiagnostic}s. Parse-only — no resolution.
+ *
+ * Diagnostics policy (#780): a single malformed region reports one specific
+ * code. After a parse error fires, leftover-token noise from the same
+ * region is suppressed (see {@linkcode expectEof}) rather than re-flagged
+ * as a redundant or contradictory second code, and a reserved character
+ * stops structural parsing outright (the lexer dropped the char, so the
+ * token stream is corrupt). Orthogonal region checks — the nav tail's
+ * reserved-char scan, the missing event dictionary (UXIL-006) — still
+ * co-report. Editors reparse per keystroke, so within a region errors
+ * surface progressively.
  */
 import type {
   ChildSurfaceDecl,
@@ -39,17 +49,30 @@ class Cursor {
   }
 }
 
-/** Push UXIL-002 for a reserved `?`/`#` character anywhere in `source`. */
-function scanReservedChars(source: string, diags: UxilDiagnostic[]): void {
+/**
+ * Push UXIL-002 for a reserved `?`/`#` character anywhere in `source`.
+ * Returns whether one was found — callers stop structural parsing on a hit,
+ * because the lexer silently drops the char and every diagnostic derived
+ * from the mangled token stream would be noise (#780). `baseColumn` shifts
+ * the reported column for callers scanning a slice of a larger span (the
+ * peeled `-> nav` tail), keeping positions span-relative.
+ */
+function scanReservedChars(
+  source: string,
+  diags: UxilDiagnostic[],
+  baseColumn = 0,
+): boolean {
   const idx = source.search(/[?#]/);
   if (idx >= 0) {
     diags.push(
       uxilDiagnostic("UXIL-002", { char: source[idx] }, {
         line: 1,
-        column: idx + 1,
+        column: baseColumn + idx + 1,
       }),
     );
+    return true;
   }
+  return false;
 }
 
 /** Consume `IDENT`, or push UXIL-001 describing what was expected. */
@@ -134,7 +157,13 @@ function parseKey(c: Cursor, diags: UxilDiagnostic[]): UxKey | undefined {
   return undefined;
 }
 
-/** Consume a leading `ux:` scheme if present. Returns whether it was consumed. */
+/**
+ * Consume a leading `ux:` scheme if present. Returns whether it was consumed.
+ * Known limitation (#780 case 6, S7 design doc): a surface literally named
+ * `ux` followed by `:` is greedily read as the scheme, so
+ * `parseRootDecl("ux : screen")` misparses. Revisit only if a real
+ * `ux`-named surface is ever needed.
+ */
 function consumeScheme(c: Cursor): boolean {
   if (
     c.peek().kind === "IDENT" && c.peek().value === "ux" &&
@@ -147,8 +176,14 @@ function consumeScheme(c: Cursor): boolean {
   return false;
 }
 
-/** Push UXIL-001 for any leftover tokens before EOF. */
+/**
+ * Push UXIL-001 for any leftover tokens before EOF. No-ops when a prior
+ * diagnostic exists — a specific parse error already explains the malformed
+ * tail, and re-flagging the leftovers produced the redundant second code
+ * (#780). One clear error per span; the editor reparses on the next edit.
+ */
 function expectEof(c: Cursor, diags: UxilDiagnostic[]): void {
+  if (diags.length > 0) return;
   if (!c.atEof()) {
     const t = c.peek();
     diags.push(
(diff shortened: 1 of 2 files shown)


stasson:
## /review — multi-angle review of this PR (8 finder angles → 1-vote verify)

18 candidates from 8 angles, deduped to 8 mechanisms, each verified (empirically where executable). **6 survived**; the 3 confirmed correctness findings are **fixed in the amended commit** (`f993d99` → `4560a4e`).

### Confirmed — fixed in this PR

1. **`grammar.ts` — nav reserved char silently lost on struct-part early return** (CONFIRMED, empirical). `` `foo -> media.home?x` `` returned only `UXIL-001`: the struct scan doesn't see the nav tail, and the early return never reached `parseUxRef(navSource)` — so nobody scanned the `?` (the pre-#780 full-span scan did). *Fix:* every return path now scans the peeled nav tail (`scanNavReserved`); pinned by the "reserved char in the nav survives a malformed struct part" test (`["UXIL-001","UXIL-002"]` — orthogonal regions co-report).
2. **`grammar.ts` — one-error policy didn't cover direct structural pushes** (CONFIRMED, empirical). Only `expectEof` was guarded: `parseRootDecl("media.home?x")` → `[UXIL-002, UXIL-004]`, and `` `/play?x : activate` `` → `[UXIL-002, UXIL-001]` where the UXIL-001 ("expected ':' before the verb set") is itself contradictory — a `:` IS present; the lexer had dropped the `?` and orphaned `x`. The exact #780 defect class, half-fixed. *Fix:* a reserved-char hit now stops structural parsing outright (the token stream is corrupt); `scanReservedChars` returns a found flag and all four parsers bail. Matches #780's own suggested direction ("stop the parser").
3. **`grammar.ts` — surviving nav `UXIL-002` column regressed to nav-relative** (CONFIRMED, empirical). Pre-PR, the (duplicated) full-span copy carried the span-relative column; the PR kept only the `parseUxRef(navSource)` copy at column 11 where the `?` sits at span column 31 — editor squiggles would land inside the verb region. *Fix:* nav diagnostics are shifted by `navOffset` at the splice point, so all nav positions (not just UXIL-002) are now span-relative; test asserts column 31.
4. **`grammar.ts` — `expectEof` statement order silently load-bearing** (CONFIRMED). The suppression guard means `expectEof` must run before the nav / UXIL-006 pushes; nothing documented it and no test pinned it — a "validate at the end" refactor would silently kill leftover-token reporting for every bullet missing its prose. *Fix:* call-site comment + pinning test (`` `/play : activate junk` `` → `["UXIL-001","UXIL-006"]`).
5. **`grammar.ts` — module-header policy comment overstated the invariant** (CONFIRMED). "One code per input/span" is false as written — orthogonal region checks (UXIL-006, nav scan) legitimately co-report; an S9 bridge author handling `diags[0]` would drop real defects. *Fix:* reworded to per-region suppression with an explicit orthogonal-co-report carve-out.

### Plausible — tracked outside this PR

6. **S8 gating contract discoverability** (PLAUSIBLE). The "gate uxRegistry registration on a clean parse, never on the partial decl alone" obligation lives in `parseRootDecl`'s doc comment; the S7 design doc §9 handoff contract doesn't mention partial decls, and `RootDecl.kind: string` can't distinguish `""` from a real kind. Resolution: noted on the S8 story #726 rather than amending the landed S7 wip doc.

### Refuted (for the record)

- *`parseRootDecl("media.home screen")` still says "missing kind" though a kind is present* — REFUTED: UXIL-004's message embeds the expected `'ux:surface : kind'` shape (self-explaining, not contradictory), and `classifyUxilForm` never routes a colon-less span to `parseRootDecl`, so the input is unreachable through the S8 pipeline.
- *Two `UXIL-004` push sites should collapse* — REFUTED: multi-site pushes of one code are this file's idiom (UXIL-008 has six), and the proposed collapse would silently accept `media.home screen` as a *valid* decl with `kind:"screen"` and no diagnostic.

Issue #771, opened 2026-07-04
Federated upstream slice 4 — deferred review follow-ups (altitude, efficiency, polish)


stasson:
Deferred, non-blocking findings from the reviews of PR #765 (federated upstream
slice 4, epic #741). The two CONFIRMED bugs (`MSL-L004` on upstream targets;
coverage `withSatisfies` leaf gate) were fixed in the PR (`fa7d968`); these are
the follow-ups that were triaged as deferrable.

## Altitude / maintainability

- **Partition entries once instead of scattering `isUpstreamEntry` across ~10
  emit loops.** The validation-exemption is a hand-added `if
  (isUpstreamEntry(entry)) continue;` at every per-entry emit loop across
  `validator/mod.ts`, `pipeline.ts`, `trace_types.ts`, `discipline.ts`,
  `types.ts`, and `typl/validator.ts`. This is structurally fragile: PR #765
  itself proved it — `validateTypl` silently leaked diagnostics against
  read-only upstream entries until the whole-branch review caught it, and the
  `MSL-L004`-on-upstream-targets bug was the same class (a check that forgot
  the guard). Refactor the pipeline to partition once at the top —
  `emittableEntries` (project + profile-corpus) vs `resolutionEntries` (all,
  incl. upstream) — feed emit loops the former and resolution-map builders the
  latter, so a future validation stage can't re-introduce the gap.

- **Consolidate the four feed sites' lockfile→upstream hydration.**
  `compileProject` (cli/helpers.ts), `check.ts`, `seedUpstreamCorpus`
  (lsp/server.ts), and `loadLockedUpstreams` (mcp/project.ts) each
  independently read `markspec.lock` → `parseLockfile` →
  `upstreamRefsFromLockfile` → `loadUpstreamCorpus` with the same
  empty-refs short-circuit and cold-cache soft-fail posture. `loadProjectUpstreams`
  covers only 2 of the 4 because it lives in `cli/helpers.ts` with a hardcoded
  Deno reader. Move it to `core/upstream/` taking an injected `readFile:
  ReadFile` (like `loadUpstreamCorpus`) so all four surfaces share one
  function and their soft-fail/diagnostic semantics can't drift.

## Efficiency

- **LSP `reloadProfile` over-reacts to `markspec.lock` changes.** A bare
  `markspec lock` re-run now triggers a full profile-chain re-resolve, a
  re-seed of the entire delivered corpus, and a `markspec/profileChanged`
[...]

PR #797, opened 2026-07-05
refactor(core): consolidate lockfile→upstream hydration (#771)

stasson:
Part of #771 — the two "altitude / maintainability" feed-site items plus the two adjacent efficiency items. (The partition-once validator refactor and the polish batch from #771 are follow-up PRs; this does not close the issue.)

## What

**Consolidate the four feed sites' lockfile→upstream hydration.** `loadProjectUpstreams` moves from `cli/helpers.ts` (hardcoded Deno reader, covering 2 of 4 sites) to `core/upstream/project.ts` with an injected `ReadFile` — the same purity rule as its siblings `loadUpstreamCorpus` / `loadDeliveredCorpus`. All four surfaces now share one function, so the empty-refs short-circuit and cold-cache soft-fail posture cannot drift:

- `cli/helpers.ts` `compileProject`
- `cli/commands/check.ts` (composite gate)
- `lsp/server.ts` `seedUpstreamCorpus`
- `mcp/project.ts` `loadLockedUpstreams`

**`report.ts` double-load.** `compileProject` now returns the `ProjectConfig` it already resolved; `report` uses it instead of re-loading `project.yaml` (additive return-type change — the other 12 callers are untouched).

**LSP `markspec.lock`-only routing.** A bare `markspec lock` re-run no longer triggers a full profile-chain re-resolve, a delivered-corpus re-seed, and a `markspec/profileChanged` push. The watched-files handler classifies the debounce window via the new pure `lsp/watched_files.ts` (`isLockfileOnlyChange`): a lock-only window reloads the lockfile, re-seeds the upstream corpus, and republishes diagnostics; any profile-file change inside the same window escalates to the full reload.

## Tests

- New unit tests: `core/upstream/project_test.ts` (no-lockfile / no-snapshot-rows short-circuits never touch `readFile`; snapshot row delegates to `loadUpstreamCorpus` and surfaces `UPSTREAM-SNAPSHOT-002`), `lsp/watched_files_test.ts` (lock-only / profile / mixed / empty batches).
- Existing coverage exercised: federated e2e (16 tests), report e2e, MCP suite (196), LSP suite (397). `just build` green.

## Docs

ADR-031 §D4 + as-built inventory now name the shared loader. The implementation plan is gardened to `docs/archive/plans/`; `docs/wip/` carries nothing from this story.

### DIFF
--- a/docs/architecture/adr-031-federated-upstream-resolution.md
+++ b/docs/architecture/adr-031-federated-upstream-resolution.md
@@ -107,15 +107,18 @@ export type EntryOrigin =
 `loadUpstreamCorpus` (`core/upstream/mod.ts`) — the sibling of ADR-030's
 `loadDeliveredCorpus`, same purity rules (injected `readFile`, no I/O of its
 own) — hydrates each cached snapshot into `Entry[]` and stamps the origin.
-Upstream entries seed at the same three sites as the ADR-030 corpus, after it,
-before project files: the CLI compiler, the LSP `seedUpstreamCorpus()`, and the
-MCP server via the shared compile cache. Read-only is structural, not a runtime
-flag — upstream entries have no local `.md` file, so `fmt`/`insert` cannot touch
-them and rename is refused by the existing `origin` guard. The one new rule:
-**go-to-definition is a no-op** for an upstream entry
-(`resolveNavigableLocation` returns `null`) — its `location.file` is an
-upstream-repo path that does not exist locally. Hover, `show`, `context`, and
-completion work from the hydrated entry.
+`loadProjectUpstreams` (`core/upstream/project.ts`) wraps the
+lockfile→refs→hydration chain — including the no-lockfile and empty-refs
+short-circuits — behind one call, so every feed surface shares the same
+soft-fail and diagnostic semantics (#771). Upstream entries seed at the same
+three sites as the ADR-030 corpus, after it, before project files: the CLI
+compiler, the LSP `seedUpstreamCorpus()`, and the MCP server via the shared
+compile cache. Read-only is structural, not a runtime flag — upstream entries
+have no local `.md` file, so `fmt`/`insert` cannot touch them and rename is
+refused by the existing `origin` guard. The one new rule: **go-to-definition is
+a no-op** for an upstream entry (`resolveNavigableLocation` returns `null`) —
+its `location.file` is an upstream-repo path that does not exist locally. Hover,
+`show`, `context`, and completion work from the hydrated entry.
 
 ### D5 — Keep one flat display-ID space; collisions are hard diagnostics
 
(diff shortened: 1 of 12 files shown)


stasson:
## /review — 8-angle finder pass + adversarial verification

**Overview.** The PR does what it says: `loadProjectUpstreams` moves to `core/upstream/project.ts` behind the `core/mod.ts` boundary with an injected `ReadFile`, all four feed surfaces call it, `compileProject` returns its already-resolved config to kill `report`'s double-load, and the LSP routes `markspec.lock`-only watcher windows to a cheap upstream re-seed. Cross-file tracing found all 13 `compileProject` callers unaffected (additive return field), no stale imports, no import cycles, lazy-loading preserved, and no AGENTS.md violations. Three findings survived verification — none blocking, all with one-line-ish remedies.

### Findings (most severe first)

1. **`packages/markspec/lsp/server.ts:789` — `reloadWatchedFiles` is not serialized against itself, so a lock-only re-seed can interleave with an in-flight full profile reload.** *(PLAUSIBLE; pre-existing class, widened by the second branch.)* `debounce` fires `fn(...)` and discards the promise, so a second 500 ms window expiring while `reloadProfile` is still awaiting starts a concurrent run. The verifier traced every await boundary: the feared "index vs `corpusFilePaths` divergence" cannot happen (each insert/clear pairs index + guard-set mutation synchronously), but two real windows remain: (a) a lock-only `publishAllDiagnostics()` can fire while a suspended `reloadProfile` has cleared upstream entries — the exact invariant the `seedUpstreamCorpus` doc comment calls load-bearing ("safe only because seedUpstreamCorpus always runs immediately afterward"); (b) last-writer-wins staleness when an in-flight hydration pinned to lock v1 resolves after a newer run seeded v2. Pre-PR, two overlapping `reloadProfile` runs could already interleave; the dispatcher adds the full ∥ lock-only pairing. A run-next-after-current promise chain in `reloadWatchedFiles` closes the whole class, including the pre-existing half.

2. **`packages/markspec/lsp/server.ts:790` — `pendingProfileFileChange` is consumed before the `await`, losing the old self-healing retry.** *(CONFIRMED; low severity.)* Pre-PR, any later watched event — including a lock-only one — re-ran the full `reloadProfile`, healing a previously failed/stale profile reload. Now `reloadProfile` swallows its own failure (`catch` → `console.warn`) after the flag is already cleared, so a subsequent `markspec lock` event takes only the cheap branch and the editor keeps stale `profile`/`cachedProfileResponse`/delivered corpus until the next profile-file save. Realistic triggers are narrow (every sub-step is designed not to throw), but the hardening is one line: clear the flag only after `reloadProfile` completes, or re-set it in a catch.

3. **`packages/markspec/core/upstream/project.ts:10` — the new doc comment overstates the consolidation: "so the four surfaces' soft-fail and diagnostic semantics cannot drift" — while `markspec.lock` read+parse still diverges one layer up.** *(CONFIRMED; pre-existing divergence, doc-only fix.)* A malformed lockfile still behaves four ways: MCP discards `parseLockfile` diagnostics, the LSP logs them and keeps the *previous* lockfile, `compileProject` drops them, `check` routes them through the MSL-L212 gate. Each caller has a verified reason to own its read+parse (L212 gate + edge ledger, module-scoped state for `markspec/version`, mtime for `isStale()`), so the seam is deliberate — the doc comment should say so and scope its "cannot drift" claim to the hydration step. Candidate wording for a follow-up or amend.

### Verified and not carried forward

- Lock-only branch lacking a try/catch: factually true, but `publishAllDiagnostics` is invoked unwrapped at `onDidSave` and in `debouncedValidateAll` too, and the global handlers log and keep the server alive — wrapping only this site would be the inconsistency.
- `isLockfileOnlyChange` keying on basename while the glob is `**/markspec.lock`: a nested/fixture lockfile edit now takes the *cheaper* path than pre-PR's full reload — improvement with residual inefficiency, optional tightening only.
- `seedDeliveredCorpus`/`seedUpstreamCorpus` byFile-seeding duplication: pre-existing, and the one divergent line (dual tracking sets) is intentionally visible per the doc comment — defer.
- `seedUpstreamCorpus`'s internal catch as "dead code": refuted — the loader's no-throw contract covers cache misses, not the index-update body; keeping it is sound.

PR #798, opened 2026-07-05
refactor(core): partition upstream-exempt entries once in the validator (#771)

stasson:
Part of #771 — PR 2 of 3: the partition-once altitude refactor. (PR 1 was #797, the feed-site consolidation; the polish batch remains.) Does not close the issue.

## What

**One partition instead of eleven guards.** `emittableEntries` (`core/model`, beside `isUpstreamEntry`) is now the single named concept for "upstream entries are read-only graph citizens: never emit against them, always resolve to them" (ADR-031 §4.7). The orchestrators partition once per entry-set generation — `validate()` at its top; `runPipeline()` for the input generation and again for the post-classification generation (Stage 2 produces new `Entry` objects) — and feed emit loops the emittable list while resolution maps keep the full list.

Collapsed guard sites: `validator/mod.ts` (`checkStructural`, `checkReferences` ×2), `pipeline.ts` (Stages 1.5, 2.4, 3, 4), `discipline.ts`, `trace_types.ts`, `typl/validator.ts` (its typl-inert boundary now expressed via the shared helper). PR #765 proved this scatter fragile — `validateTypl` and the `MSL-L004` target check both shipped without the guard until review caught them; a future stage now gets the partition by construction.

**Deliberately kept:** `classifyEntriesStage`'s upstream branch (a transform, not an emit loop — pass-through preserves resolution targets and first-entry-wins order; upstream `type` comes from its own compile), and `traceability.ts`'s `MSL-L004` target-side guard (different mechanism, #765's fix).

**Consistency crumbs from #771:**

- `MSL-R083` gains the target-side upstream exemption in both the `TRACE_RULES` and polymorphic `Caused-by` loops — the target-side twin of #765's `MSL-L004` fix. Behavior change: a project entry's trace link to an upstream entry whose display-ID-inferred core type is incompatible no longer errors.
- `computeCoverage` uses the shared `isUpstreamEntry` predicate (now a type predicate, so the narrowing carries `origin.upstreamId`).

## Tests

- **Upstream-silence anchor** (`pipeline_test.ts`): one upstream entry baiting a rule in every emit stage (missing Id, empty title, unknown attribute, unresolvable Supersedes/References, bad Type, unknown Discipline, retired-target severity) must produce zero diagnostics through `runPipeline` while remaining a live resolution target. This is the #765 regression net — it pins the partition, not the implementation, and passed both before and after the refactor.
- **R083 target exemption** (`trace_types_test.ts`): red before the fix (R083 fired on an upstream `Component` target), green after. Direct-call tests compose `emittableEntries` the way the pipeline does, so they keep exercising the production caller contract.
- `emittableEntries` unit test (order-preserving, corpus entries stay emittable).
- Full suite: 2248 core tests, `just build` green.

## Docs

ADR-031 §D4 + as-built inventory record the partition (`isUpstreamEntry` remains the target-side and reporter predicate). Plan gardened to `docs/archive/plans/`.

### DIFF
--- a/docs/architecture/adr-031-federated-upstream-resolution.md
+++ b/docs/architecture/adr-031-federated-upstream-resolution.md
@@ -103,11 +103,13 @@ export type EntryOrigin =
 ```
 
 `formatEntryOrigin` renders the upstream badge as `<upstreamId>@<version>` (e.g.
-`product@v2.1.0`); `isUpstreamEntry` gates the validator and lint loops.
-`loadUpstreamCorpus` (`core/upstream/mod.ts`) — the sibling of ADR-030's
-`loadDeliveredCorpus`, same purity rules (injected `readFile`, no I/O of its
-own) — hydrates each cached snapshot into `Entry[]` and stamps the origin.
-`loadProjectUpstreams` (`core/upstream/project.ts`) wraps the
+`product@v2.1.0`). `emittableEntries` (`core/model`) partitions the validators'
+emit side once per run — upstream entries never enter an emit loop but stay in
+every resolution map (#771); `isUpstreamEntry` remains the target-side and
+reporter predicate. `loadUpstreamCorpus` (`core/upstream/mod.ts`) — the sibling
+of ADR-030's `loadDeliveredCorpus`, same purity rules (injected `readFile`, no
+I/O of its own) — hydrates each cached snapshot into `Entry[]` and stamps the
+origin. `loadProjectUpstreams` (`core/upstream/project.ts`) wraps the
 lockfile→refs→hydration chain — including the no-lockfile and empty-refs
 short-circuits — behind one call, so every feed surface shares the same
 soft-fail and diagnostic semantics (#771). Upstream entries seed at the same
@@ -379,8 +381,9 @@ are byte-reproducible across machines from `(tree, markspec version)`.
 - [ADR-032](./adr-032-process-profile-boundary.md) — settles the sibling
   `process:` ↔ profile-activation question this design scoped out (§9 #2).
 - As-built: `core/model/mod.ts` (`EntryOrigin`, `formatEntryOrigin`,
-  `isUpstreamEntry`, `ProjectRef`, `ProjectConfig`), `core/upstream/mod.ts`
-  (`loadUpstreamCorpus`), `core/upstream/refs.ts` (`upstreamRefsFromLockfile`),
+  `isUpstreamEntry`, `emittableEntries` — the #771 emit-partition helper,
+  `ProjectRef`, `ProjectConfig`), `core/upstream/mod.ts` (`loadUpstreamCorpus`),
+  `core/upstream/refs.ts` (`upstreamRefsFromLockfile`),
   `core/upstream/project.ts` (`loadProjectUpstreams`, shared by all four feed
   surfaces — #771), `core/lock/model.ts` + `serializer.ts`
   (`UpstreamDependency`, extended `UpstreamRegistry`),
--- a/docs/archive/plans/2026-07-05-771-partition-once-plan.md
+++ b/docs/archive/plans/2026-07-05-771-partition-once-plan.md
@@ -1,0 +1,563 @@
+# Partition-Once Validator Refactor (#771 PR 2) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use
+> superpowers:subagent-driven-development (recommended) or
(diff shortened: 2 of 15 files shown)


stasson:
## /review — 8-angle finder pass + adversarial verification

**Overview.** The PR replaces the ~11 scattered `isUpstreamEntry` emit-loop guards with the `emittableEntries` partition, computed once per entry-set generation in `validate()` and `runPipeline()`, adds the MSL-R083 target-side upstream exemption, and moves `computeCoverage` onto the shared predicate. The correctness angles came back clean: every removed guard maps to a pre-filtered loop over the same generation (including the subtle `finalEmittable` recompute after Stage 2.5's reassignment), `checkStructural`'s duplicate maps are membership-identical, resolution maps build from the same full lists as before, all callers of the changed signatures are updated, the type-predicate narrowing is safe at every site, and the R083 `continue`s cannot reach the deliberately-kept R084/R081/R082 loops. Four findings survived verification — all non-blocking consistency/hardening items.

### Findings (most severe first)

1. **`packages/markspec/core/lint/runner.ts:174` (also `:200`, `:239`, `core/gates/mod.ts:69`, `:139`) — the lint/gates subsystem runs a deliberately BROADER partition (`!entry.origin`: excludes profile corpus AND upstream) that is now unnamed and undocumented next to the new concept.** *(CONFIRMED.)* Two sharp edges: the new helper's doc comment commands "Adding a new validation stage? Iterate the emittable list — never the raw entry set" — which is exactly the *wrong* instruction inside lint/gates (it would re-enable prose lint on profile-corpus entries, the ADR-030 §D4 regression); and the PR deleted ADR-031's old "`isUpstreamEntry` gates the validator **and lint** loops" sentence, leaving lint's partition documented nowhere. Right-sized fix, zero behavior change: a carve-out sentence in the `emittableEntries` doc comment plus a cross-reference comment (or a named local partition) in `runner.ts`/`gates/mod.ts`.

2. **Three one-sentence documentation gaps around the partition's deliberate boundaries.** *(CONFIRMED as a cluster; one tiny `docs(core)` follow-up commit.)*
   - `pipeline.ts:100` — "Partition once" is per-orchestrator: `validate()` (Stage 1) and `validateTypl` re-partition the same generation, and the enforcement is convention + the anchor test, not structural. Soften the comment.
   - `validator/mod.ts:84` — `validate()` hands the FULL list to `validateTypl` four lines after "emit loops iterate `emittable`"; one sentence ("validateTypl applies the partition itself — registry-scope contract") closes the fix-the-inconsistency → delete-typl's-filter → leak drift path.
   - `core/model/mod.ts:494` — `emittableEntries` lives in model because moving it to `validator/` would create the typl→validator cycle the dependency-flow diagram forbids; the doc comment should record that constraint so a future "move it where it belongs" refactor doesn't fork the predicate.

3. **`packages/markspec/mcp/resources/entry.ts:43` (and `:31`) — the last two hand-rolled `origin.kind === "upstream"` checks in the tree.** *(CONFIRMED.)* Line 43's next line reads `entry.origin.upstreamId` — exactly the narrowing the new type predicate provides, and the same conversion the PR applied to the reporter twin. A tree-wide grep confirms these are the only survivors outside `core/model`. Two-line mechanical follow-up.

4. **`packages/markspec/core/validator/pipeline.ts:162` — `let finalEmittable` computed after Stage 2 and reassigned inside the Stage 2.5 branch.** *(CONFIRMED, small hardening.)* A stage added between 2.4 and 2.5 that uses the binding silently iterates the pre-normalization generation. Verified fix (b): inline `emittableEntries(finalEntries)` at Stage 2.4's loop and declare a single `const finalEmittable` after Stage 2.5 — behavior-identical, removes the mutable generation-tracking. (Option (a), moving Stage 2.4 after 2.5, is also provably safe — `inferTypeFromLateStageChain` never reads `typedAttributes` — but forces renumbering the stage narrative.)

### Verified and not carried forward

- **MSL-R085 judging an upstream target's shape: REFUTED as an inconsistency** — shape is a core authored fact preserved verbatim through hydration, which is precisely the criterion the PR uses to keep R084/R081/R082 unexempted. At most a future design question ("may `References:` cite an upstream Authored entry?"), not a defect.
- The added filter allocations (~5 O(n) passes per pipeline run) are noise next to the existing map builds and regex scans; a zero-upstream fast path would be premature.
- R083-under-ID-collision edge: when an upstream and a project entry collide on a display ID, the first-wins map picks the upstream twin and the new exemption skips R083 for links naming that ID — but the collision is already an MSL-R014 error, so nothing actionable is hidden.

PR #799, opened 2026-07-05
fix(core): federated-upstream polish — remedies, path hygiene, hardening (#771)

### LINKED ISSUE #771: Federated upstream slice 4 — deferred review follow-ups (altitude, efficiency, polish)

stasson:
Closes #771 — PR 3 of 3, the polish batch. (PR 1 = #797 feed-site consolidation; PR 2 = #798 partition-once refactor.)

## Issue items

- **`cacheRoot` path hygiene.** The cache-dir construction in `upstream_refs.ts` and `cache_check.ts` uses `join()` instead of a `${cacheRoot}/${id}` template — the concat mixed the platform separator with a literal `/`, the exact Windows drift class that failed CI twice on slices 2/4. Writer and reader (`upstreamRefsFromLockfile`, already `join()`-built) now agree byte-for-byte.
- **Consumer-actionable `MSL-R014` for upstream↔upstream collisions.** When two upstreams independently author the same display ID/Id, the consumer can rename neither read-only entry — the message now says to drop one of the colliding upstreams from `project.yaml` or coordinate an upstream rename. Mixed-tier collisions keep the rename wording (someone owns that entry).
- **Origin-flavored attribution.** A finding located in a hydrated upstream snapshot reads `from upstream X:` instead of the ADR-030 profile-corpus `delivered by X:`.
- **Test hardening:** `kind:"dependency"` fixture in `refs_test.ts`; the `MSL-T014` message pinned by exact equality (was `.includes`); URI trace targets with upstreams declared stay silent (the URI short-circuit precedes federation); new e2e — a cold/missing upstream cache surfaces `UPSTREAM-SNAPSHOT-002` on stderr while `show` still exits 0 (soft-fail through `compileProject`).
- **Deliberately not fixed (noted limitation):** MCP lockfile staleness stays mtime-only. SHA-gating the lock read is disproportionate to the risk of a same-mtime lockfile rewrite; recorded here so closing the issue doesn't silently drop the item.

## #797 review findings applied

- LSP watched-file reloads are **serialized** (run-next-after-current promise chain) — a debounce window expiring while a reload is in flight can no longer interleave the corpus seeders' shared-set mutations (this also closes the pre-existing full∥full pairing); failures land in a scoped warning instead of the global handler.
- A **failed profile reload keeps the escalation flag set**, so the next watched event of any kind retries the full reload — restoring the self-healing property the dispatcher split had dropped.
- `loadProjectUpstreams`' doc scopes its "cannot drift" claim to the hydration step and records the deliberate per-caller lock read+parse seam.

## #798 review findings applied

- `emittableEntries` doc gains the **lint/gates carve-out** (their broader `!entry.origin` partition is deliberate — profile corpus is prose-lint-exempt per ADR-030 §D4; unifying would regress) and the model-placement rationale (a `validator/` home would cycle via typl); cross-reference comments at all three lint/gates filter sites.
- `pipeline.ts` "Partition once" comment corrected to per-orchestrator; `finalEmittable` is a single `const` derived after Stage 2.5 (Stage 2.4 partitions inline) — no more mutable generation-tracking; `validate()` documents why `validateTypl` receives the full list.
- `mcp/resources/entry.ts` uses the shared `isUpstreamEntry` type predicate — the last two hand-rolled `origin.kind === "upstream"` checks in the tree.

## Docs

ADR-031 §D4 records the broader lint/gates partition next to the `emittableEntries` sentence. Plan gardened to `docs/archive/plans/`.

### DIFF
--- a/docs/architecture/adr-031-federated-upstream-resolution.md
+++ b/docs/architecture/adr-031-federated-upstream-resolution.md
@@ -106,21 +106,24 @@ export type EntryOrigin =
 `product@v2.1.0`). `emittableEntries` (`core/model`) partitions the validators'
 emit side once per run — upstream entries never enter an emit loop but stay in
 every resolution map (#771); `isUpstreamEntry` remains the target-side and
-reporter predicate. `loadUpstreamCorpus` (`core/upstream/mod.ts`) — the sibling
-of ADR-030's `loadDeliveredCorpus`, same purity rules (injected `readFile`, no
-I/O of its own) — hydrates each cached snapshot into `Entry[]` and stamps the
-origin. `loadProjectUpstreams` (`core/upstream/project.ts`) wraps the
-lockfile→refs→hydration chain — including the no-lockfile and empty-refs
-short-circuits — behind one call, so every feed surface shares the same
-soft-fail and diagnostic semantics (#771). Upstream entries seed at the same
-three sites as the ADR-030 corpus, after it, before project files: the CLI
-compiler, the LSP `seedUpstreamCorpus()`, and the MCP server via the shared
-compile cache. Read-only is structural, not a runtime flag — upstream entries
-have no local `.md` file, so `fmt`/`insert` cannot touch them and rename is
-refused by the existing `origin` guard. The one new rule: **go-to-definition is
-a no-op** for an upstream entry (`resolveNavigableLocation` returns `null`) —
-its `location.file` is an upstream-repo path that does not exist locally. Hover,
-`show`, `context`, and completion work from the hydrated entry.
+reporter predicate. The prose-lint runner and the check gates apply a
+deliberately broader origin-based partition — profile-corpus and upstream prose
+are both lint-exempt (ADR-030 §D4). `loadUpstreamCorpus`
+(`core/upstream/mod.ts`) — the sibling of ADR-030's `loadDeliveredCorpus`, same
+purity rules (injected `readFile`, no I/O of its own) — hydrates each cached
+snapshot into `Entry[]` and stamps the origin. `loadProjectUpstreams`
+(`core/upstream/project.ts`) wraps the lockfile→refs→hydration chain — including
+the no-lockfile and empty-refs short-circuits — behind one call, so every feed
+surface shares the same soft-fail and diagnostic semantics (#771). Upstream
+entries seed at the same three sites as the ADR-030 corpus, after it, before
+project files: the CLI compiler, the LSP `seedUpstreamCorpus()`, and the MCP
+server via the shared compile cache. Read-only is structural, not a runtime flag
+— upstream entries have no local `.md` file, so `fmt`/`insert` cannot touch them
+and rename is refused by the existing `origin` guard. The one new rule:
+**go-to-definition is a no-op** for an upstream entry
+(`resolveNavigableLocation` returns `null`) — its `location.file` is an
+upstream-repo path that does not exist locally. Hover, `show`, `context`, and
+completion work from the hydrated entry.
 
 ### D5 — Keep one flat display-ID space; collisions are hard diagnostics
 
--- a/docs/archive/plans/2026-07-05-771-polish-plan.md
+++ b/docs/archive/plans/2026-07-05-771-polish-plan.md
@@ -1,0 +1,109 @@
+# #771 PR 3 — Polish Batch Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:executing-plans.
+> Steps use checkbox (`- [ ]`) syntax.
+
+**Goal:** Close #771 with the remaining polish: cacheRoot concat normalization,
+a consumer-actionable MSL-R014 remedy for upstream↔upstream collisions, the
+issue's test-hardening list, and the accumulated non-blocking review findings
+from PRs #797 and #798. Deliberately NOT fixed: MCP lockfile mtime-only
+staleness (noted limitation — SHA-gating the lock read is disproportionate;
+recorded in the PR body).
+
+**Architecture:** All items are local; no signatures change except
+`reloadProfile(): Promise<boolean>` (LSP-internal). One squashed commit.
+
+## Task 1 — cacheRoot normalize-at-construction (#771 issue item)
+
+- `core/lock/upstream_refs.ts:258`: ``const dir = `${opts.cacheRoot}/${id}`;`` →
+  `const dir = join(opts.cacheRoot, id);` (add `join` to the `@std/path` import
+  if absent).
+- `core/lock/cache_check.ts:51`: same change with `cacheRoot`/`upstream.id` (add
+  import).
+- No third site exists (grep verified; `core/upstream/refs.ts` already joins).
+- Verify: `deno test packages/markspec/core/lock/` — behavior-preserving; the
+  writer now agrees with the join-built reader (`upstreamRefsFromLockfile`), the
+  exact Windows drift class from slices 2/4.
+
+## Task 2 — MSL-R014 upstream↔upstream remedy (TDD)
+
+- RED test in `core/validator/corpus_test.ts`: two entries with DIFFERENT
+  `kind:"upstream"` origins sharing a display ID → the R014 message must NOT say
+  "one must rename its entry" (consumer can rename neither); must name a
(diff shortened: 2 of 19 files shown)


stasson:
## /review — 6-angle finder pass; confirmed findings FIXED IN-BRANCH (amended into `5745834`)

**Overview.** The PR delivers the #771 polish batch: `join()`-built cache dirs, the consumer-actionable MSL-R014 remedy + `from upstream` attribution, the issue's test-hardening list, and the #797/#798 review follow-ups (reload serialization, escalation-flag self-healing, partition documentation, the last hand-rolled predicates). The finder pass verified the mechanics sound — chain doesn't grow unboundedly, a rejection can't block the next run, the flag snapshot/restore form is load-bearing (clear-on-success would clobber mid-flight escalations), fixture keys survive the `join()` change, the pinned T014 string is byte-identical to its producer — and surfaced five actionable findings plus three low-severity residuals. Since the review ran pre-merge on my own PR, the five actionable ones were **fixed in the amended commit** rather than deferred:

### Fixed in-branch (`outcome: fixed`)

1. **Every `.md` save escalated to a full profile reload.** The shipped VS Code extension synchronizes `**/*.md` (`extension.ts:148`), delivering every markdown change to `onDidChangeWatchedFiles` regardless of the server's 3-file dynamic registration — `isLockfileOnlyChange` classified them as profile-affecting, so the expensive branch (profile re-resolve + both corpus re-seeds + a second `validateAll` on top of `onDidSave`'s) ran 500 ms after every save, and the lock-only fast path was unreachable in the real client. Pre-existing in class, but this dispatcher is what the PR series hardened. Fix: `relevantWatchedUris` filters to the three owned basenames before any routing (+ unit tests).
2. **`MSL-R014` inter-tier remedy generalized.** The PR fixed the unactionable "one must rename" wording for upstream↔upstream — but profile↔profile and mixed collisions have the identical defect: the consumer reading the diagnostic owns neither read-only entry. The remedy is now consumer-actionable for every origin pairing (drop a delivered profile from `.markspec.yaml` / an upstream from `project.yaml`, or coordinate with the author); project↔origin keeps "rename this entry".
3. **Reload failures vanished from the event log.** The chain's `.catch` swallowed rejections before the global `unhandledrejection` handler that used to `logEvent` them into `.markspec/lsp.log` — an unintended analytics regression of the "scoped warning" change. The catch now writes the event log too.
4. **Queued chain link could run after shutdown.** `debounced.cancel()` only clears the un-fired timer; a link queued behind an in-flight reload started after `onShutdown`'s final log flush (post-shutdown disk writes + wire traffic). A `shuttingDown` gate at the top of `reloadWatchedFiles` closes it.
5. **The attribution test used `MSL-R014` as its vehicle** — a composition production never takes (collisions are appended *after* the downgrade pass at every call site and stay hard errors per ADR-031 D5); the test pinned a softening that would be a bug if ever reachable. Swapped to a generic per-entry code.

### Deferred → #800 (low severity, verified)

- Initial indexing runs outside the reload chain; a profile/lock change during the init window can still interleave (much narrowed by fix 1).
- `check --strict` would re-promote `from upstream`-attributed warnings (exemption set is delivered-corpus-only) — near-unreachable post-#798.
- Git dependencies badge as `<id>@unversioned` despite `resolved` carrying a precise pin.

### Verified clean / refuted

- No caller breakage from `reloadProfile(): Promise<boolean>` (single caller); no consumer parses the `delivered by ` prefix; `attributeCorpusDiagnostics` signature unchanged for all four call sites; no snapshot or e2e asserts the old R014 wording; conventions (library boundary, dependency flow, docs, Node compat) all pass; the promise chain releases settled links (no memory growth); the extra Stage-2.4 filter pass is noise.

All suites re-run green after the fixes (validator+LSP 768, federated e2e 7, watched-files 6).

Issue #72, opened 2026-03-23
Multi-book support


stasson:
## Context

MarkSpec projects publish multiple independent books from a single repository (e.g. `docs/spec/` and `docs/guide/`). Each book has its own `SUMMARY.md` and builds as a standalone site. Currently we use mdBook as an interim renderer, but the `markspec book build` command needs native multi-book support.

## Stories

- [ ] #73 — Parse SUMMARY.md into a chapter tree
- [ ] #74 — Generate static HTML site from parsed book
- [ ] #75 — Build multiple books from a single repository
- [ ] #76 — Live preview server for book development
- [ ] #77 — Replace mdBook with native book renderer
- [ ] #78 — Generate Pages landing page linking all books
- [ ] #79 — Link external API docs (rustdoc, Dokka) from landing page

PR #805, opened 2026-07-05
Restore mdBook as the interim book renderer

### LINKED ISSUE #804: Bring back mdBook for site chrome until native book renderer reaches parity
## Context

Epic: #72

Issue #77 replaced mdBook with the native `markspec book build` renderer for
the 4 published books (guide, spec/language, spec/typography, spec/model).
The cutover (landed via #762 on 2026-07-04) dropped mdBook's built-in site
chrome — sidebar/TOC navigation, full-text search, syntax-highlighted code
fences, a print stylesheet, and the light/dark theme toggle — without a
native replacement. Epic #72 already frames mdBook as "the interim renderer"
pending native multi-book support, so this issue restores that interim state
until the native renderer actually has the missing chrome.

## What's changing

- Restore the 4 deleted `book.toml` files (guide, spec/language,
  spec/typography, spec/model) verbatim.
- Point `.github/workflows/pages.yaml` and `justfile`'s `book`/`book-dev`
  targets back at `mdbook build`/`mdbook serve`.
- Leave `packages/markspec/book/`, `cli/commands/book.ts`, and
  `tests/e2e/book_build_test.ts` untouched — the native renderer keeps
  working and keeps its tests, it's just not wired to the public Pages site
  for now.

No published book currently has a live (non-fenced) entry block, so nothing
visually regresses in the reverse direction — entry-block/pill/alert/caption
styling isn't exercised by any of the 4 books today.

## Exit criteria (when this issue can close and a #77-style native cutover can be re-attempted)

`markspec book build` needs all of:

- [ ] Sidebar / TOC navigation generated from `SUMMARY.md`
- [ ] Full-text search across chapters
- [ ] Syntax-highlighted code fences
- [ ] Print-friendly stylesheet ("print entire book" view)
- [ ] Light/dark (or multi-theme) toggle

## Non-goals

- Not deleting or reverting the native renderer's code or tests.
- Not building a bridge to preserve entry-block coloring under mdBook — no
  published book currently has live entries, so this would be speculative.

stasson:
Closes #804.

## Summary

- Restores mdBook as the renderer for all 4 published books (guide,
  spec/language, spec/typography, spec/model), bringing back sidebar/TOC
  navigation, full-text search, syntax-highlighted code fences, a print
  stylesheet, and a light/dark theme toggle — all dropped silently when
  issue #77 cut the site over to a native `markspec book build` renderer
  that never grew this chrome. Epic #72 already framed mdBook as "the
  interim renderer" pending native multi-book support, so this reverts the
  premature part of that cutover, not the whole epic.
- The native renderer's code (`packages/markspec/book/`,
  `cli/commands/book.ts`) and its tests are left untouched — just no longer
  wired to the public Pages site. `docs/architecture/adr-033-mdbook-interim-restoration.md`
  records the decision and its exit criteria; #804 is closed by this PR
  and tracks bringing native back once it has nav/search/syntax-highlighting/
  print/theme-toggle parity.
- Found and fixed one factual error along the way: the ADR initially claimed
  mdBook copies `additional-css` into each book's build directory. It
  doesn't — every page links out via a relative `../` path to a single
  shared `_site/theme/markspec.css`, which the surviving top-level copy step
  in `justfile`/`pages.yaml` provides. That step is now load-bearing for all
  4 books' styling, not just the landing page, so both files got a guard
  comment pointing back at the ADR.

## Test plan

- [x] `mdbook build` on all 4 books individually — exit 0, no warnings, and
      chrome assets (`<nav`, `sidebar`, `searchindex.js`, `highlight.js`,
      `print.html`, `theme-list`) verified present in the output.
- [x] `just book` end-to-end (tokens → 4 mdbook builds → Typst cheat sheet →
      theme/schemas/landing-page copy) — exit 0.
- [x] `just build` (lint + full test suite + typecheck + compile) — exit 0,
      confirming the native renderer's own code/tests are unaffected.
- [x] Independently re-verified the CSS-link mechanism (no file copied by
      mdBook; relative-path link resolves to the shared top-level copy) after
      the final review flagged the ADR's original claim as incorrect.
- [x] Every task in the implementation plan
      (`docs/archive/plans/2026-07-05-restore-mdbook-plan.md`) passed an
      independent task-scoped review; the whole-branch diff passed a final
      review with one Important finding, fixed and re-verified clean.


stasson:
## Code review

Ran an 8-angle review (correctness × 4, cleanup × 3, conventions × 1) over this
PR's diff, verified each surviving candidate independently, and confirmed two
findings with live empirical checks. Full methodology: 22 raw candidates → 2
survive verification.

### Finding 1 — `additional-css` relative path is wrong for the real deployed
site; `markspec.css` 404s in production

**Files:** `docs/guide/book.toml:11`, `docs/spec/language/book.toml`,
`docs/spec/typography/book.toml`, `docs/spec/model/book.toml`

I built the site, copied the output into a directory tree simulating this
repo's actual GitHub Pages routing (confirmed via
`gh api repos/driftsys/markspec/pages` → project page at
`https://driftsys.github.io/markspec/`), and served it with a local HTTP
server. Every book's stylesheet `<link>` resolves too high:

- `docs/guide/book.toml`: `additional-css = ["../../theme/markspec.css"]` (2
  levels) — but `_site/guide/` sits exactly 1 level under `_site/`, so 1 level
  is correct. From `http://.../markspec/guide/index.html`, the declared path
  resolves to `http://.../theme/markspec.css` — outside the `/markspec/` site
  root — **HTTP 404**.
- The 3 `docs/spec/*/book.toml` files declare 3 levels (`../../../`) for the
  same reason (derived from source-dir depth, not build-dir depth) — same
  404, worse offset.

The fix is the same for all 4 files: `additional-css = ["../theme/markspec.css"]`
(single `../`), since every book's `build-dir` is a direct child of `_site/`.

This bug predates this PR — it's byte-identical to the original 2026-03
mdBook config (verified against git history) — and has zero visible impact
today only because no published book currently contains a live, unfenced
entry block that needs the custom styling. But it does mean two claims this
PR makes are incomplete:
- The guard comment added to `justfile`/`pages.yaml` ("this copy is
  load-bearing for all 4 mdBook books... do not remove") is necessary but not
  sufficient — the emitted `<link>` is wrong regardless of whether the copy
  exists.
- `docs/architecture/adr-033-mdbook-interim-restoration.md`'s Decision
  section states the shared copy step makes the CSS resolve correctly for
  all 4 books; it doesn't, on the actual deployed site.

### Finding 2 — no local `mdbook` install instructions; `just book`/`just book-dev`
will fail for a contributor following the repo's own setup flow

**File:** `bootstrap` (also `README.md`, `CONTRIBUTING.md`)

`AGENTS.md`'s "Post-clone setup" says to run `./bootstrap`; that script only
installs `git-std` and tree-sitter grammars, never `mdbook`. This PR's
`justfile` changes make `just book`/`just book-dev` hard-require the
`mdbook` binary. A contributor who clones, runs `./bootstrap` per the
documented flow, then tries `just book` to preview docs locally gets
`mdbook: command not found` with no pointer anywhere in the repo to fix it.

This is a pre-existing gap (verified via git history: `bootstrap` never
installed mdbook even during its original 2026-03–2026-04 usage) —
reactivated, not newly introduced, by this PR. `adr-033` names the
dependency explicitly ("mdBook must be installed locally") without closing
the gap.

### Investigated and refuted

- **MSL-K001 slug-collision check loss under mdBook** — reverting from
  `markspec book build` to `mdbook build` drops the native renderer's
  chapter-slug-collision check (`packages/markspec/book/mod.ts`). Verified
  this is moot: mdBook mirrors the source directory tree rather than
  flattening it (confirmed via an actual `mdbook build docs/guide` and
  inspecting `_site/guide/recipes/*.html`), so the specific collision class
  MSL-K001 exists for — two different nested source paths hyphen-flattening
  to the same output file — structurally cannot occur under mdBook's
  routing. No equivalent check is needed.
- Several duplication/verbosity observations (4 similar `book.toml` files;
  the ADR restating content from the archived design spec) were also
  surfaced but reflect the explicitly-chosen plain-revert approach (see the
  design spec's Alternatives Considered) rather than defects.

stasson:
Both findings fixed in `0a956fa`.

**Finding 1 (CSS path 404):** All 4 `book.toml` files now declare
`additional-css = ["../theme/markspec.css"]` (uniform single `../`, since
every book's `build-dir` sits exactly one level under `_site/`). Re-ran the
same live-deployment simulation from the review — both checked URLs now
resolve `200` instead of `404`.

This surfaced a second, sharper problem the original review didn't catch:
mdBook resolves `additional-css` **twice** — once from the book's *source*
root (to locate the real file for its own internal copy step) and once from
`build-dir` (to build the emitted `<link>` href). The single corrected value
satisfies the href side but not the source-side lookup, so `mdbook build`
failed outright with `Unable to copy across static files` until I added two
committed symlinks — `docs/theme/markspec.css` and
`docs/spec/theme/markspec.css` (shared by the 3 `docs/spec/*` books) — both
pointing at the canonical `theme/markspec.css`. I independently reproduced
the failure (removed the symlink, got the exact same mdBook error) before
accepting this as necessary rather than trusting the fix's own report.
Documented in `docs/architecture/adr-033-mdbook-interim-restoration.md` and
in a comment on each `book.toml`.

**Finding 2 (no local mdbook install docs):** Added a note to `AGENTS.md`'s
"Post-clone setup" section and a `bootstrap` output line — matches this
repo's existing convention of not auto-installing optional local-preview
tools (`bootstrap` doesn't install Typst either), so this documents the gap
rather than building a new cross-platform installer.

All 4 books rebuild cleanly (verified locally again after the fix), `just
build`/`just check` pass (pre-push hook: 3906 passed, 0 failed, 3 ignored),
and the branch is pushed.

Issue #807, opened 2026-07-05
fix(core): typl cross-entry codes (TYPL-009/010/011) fire falsely on file-local check


stasson:
Found during S9 (#727) review. The uxil family now suppresses its cross-entry codes (UXIL-016/017/018) under file-local `check <file>` scope — a subset registry cannot distinguish a dangling reference from an unchecked file (same false-positive class as MSL-L006, which the pipeline already filters under `projectWide: false`).

typl has the identical latent problem but no gate: `TYPL-010`/`TYPL-011` (and `TYPL-009` declared-once across files) run from Stage-1 `validate()`, which has no `projectWide` knob, so `markspec check citing-file.md` exits 1 with a false `TYPL-011` when the declaring entry lives in an unchecked file.

## Acceptance

- `markspec check <subset>` no longer emits TYPL-009/010/011 for references whose resolution requires entries outside the subset (mechanism: either move the typl cross-entry checks behind the pipeline's `projectWide` flag, or thread scope into `validateTypl`).
- Bare `markspec check`, `compile`, and the LSP (whole-workspace) behavior unchanged.
- E2E: explicit-file check of a citing file alone exits 0; bare check still reports the dangling citation.
- uxil and typl end up with symmetric file-local scope semantics (uxil side landed in #727's story PR).

Issue #781, opened 2026-07-05
uxil parser: element-bullet diagnostic positions drop the code-span offset (blocks S9 file-anchoring)


stasson:
Follow-up from the #725 / #779 code review (S7 uxil parser).

`parseElementBullet` (in `packages/markspec/core/uxil/grammar.ts`) parses the
leading inline code span of a bullet paragraph, but `splitLeadingCodeSpan`
returns only the span's **inner text** — it discards the character offset of
where the span starts in the paragraph (leading whitespace + the backtick
delimiter(s)). Every UXIL diagnostic from an element bullet therefore reports a
column relative to the de-backticked, left-trimmed span, shifted from the true
paragraph column.

Example: `` `/play :` `` reports the empty-verb error at column 8, but the `:` is
further right in the real paragraph once the leading backtick (and any list
indent) are counted.

Why it matters: S7 positions are deliberately source-local (`line: 1`), and the
S9 (#727) bridge is meant to attach the file path + offset. But because the
span's start offset is thrown away, **S9 cannot recover the correct column for
element-bullet diagnostics** — it can only anchor to the paragraph start.

Suggested fix: have `splitLeadingCodeSpan` also return the span's start offset
within the paragraph (chars consumed before the span's inner text), and thread
that offset into the diagnostics `parseElementBullet` produces (or carry it on
the returned node) so S9 can add it when bridging. The `-> nav` sub-parse offset
should compose too.

Part of #717 (uxil epic). Should land before or with #726/#727.

Issue #728, opened 2026-07-03
story(uxil): LSP — hover, completion, go-to-declaration


stasson:
Part of #717 §B. Depends on: S8 (uxRegistry).

## Goal
LSP affordances over `uxRegistry`:
- **hover** on any `ux:` ref → declaration card (kind, verb set, states, owning entry, description);
- **completion** after `ux:`;
- **go-to-declaration** from a citation to its declaration.

## Acceptance
- Hover/completion/definition unit-tested against a fixture registry (parallel to typl LSP tests).
- Source-file doc-comment context guarded like typl.

PR #808, opened 2026-07-05
story(uxil): UXIL-0xx diagnostics family (#727)

### LINKED ISSUE #727: story(uxil): UXIL-0xx diagnostics family

stasson:
Closes #727.

S9 of the uxil epic (#717 §B): the `UXIL-0xx` diagnostics family, wired end-to-end. Design spec + plan archived under `docs/archive/` (with as-built addendum).

## What's in here

- **Profile gate** — new type-level `declares: ux-surface` manifest field (closed vocabulary; unknown value is `PROFILE-TYPE-009` at load). No designation anywhere in the chain → the family is fully inert, preserving the epic's Tier-1 opaqueness: a `.gitignore`-style prose bullet can never draw a diagnostic.
- **Family orchestrator** — `core/validator/uxil_family.ts`: gates entries by resolved type (explicit `entry.type` or `classifyEntry` fallback, so the LSP path — which never runs pipeline Stage 2 — gates identically), runs the S8 compiler over declaring entries, validates `ux:` citations corpus-wide, and emits `UXIL-023` for a root declaration outside a declaring type. Upstream entries are uxil-inert (#771 partition). Lives in `validator/` so the `pipeline → family → classifyEntry` import chain stays one-directional.
- **File-anchoring** — every uxil diagnostic now leaves assembly/validation as a core `Diagnostic` with file coordinates. Grammar diagnostics compose via the #781 paragraph-relative contract plus a new `InlineDeclaration.innerColumn` on the shared decl machinery; exact positions are pinned by tests.
- **New codes** — `UXIL-023` (declaration outside the declaring entry type), `UXIL-024` (relative ref without base — reserved with cascade suppression until the table surface lands; mapped at the resolution site, direct-tested), `UXIL-025` (`observe` on a non-visual kind), `UXIL-026` (`navigate` without a `->` target). The latter two enforce vocab flags S8 shipped unchecked.
- **Wiring** — `runPipeline` Stage 5 (→ `markspec check`) and LSP `WorkspaceIndex.validateAll`. File-local `check <file>` suppresses the cross-entry codes (`UXIL-016/017/018`) — a subset registry cannot distinguish a dangling reference from an unchecked file, mirroring the MSL-L006 pattern. typl has the same latent issue file-locally; tracked as #807.
- **Catalogue + editor** — `docs/spec/language/uxil.md` (26-code table + activation/scope semantics, in SUMMARY) and LSP `codeDescription` anchors at `https://markspec.dev/spec/uxil#uxil-0xx`. The full uxil ADR + guide chapter remain S12's scope (#730).

## Acceptance (#727)

- Every code UXIL-001..026 has a template + severity in `UXIL_CODES` and a triggering test fixture (001–008 grammar; 009–022 validator/assemble; 023/025/026 new fixtures; 024 direct factory + suppression tests, documented reserved).
- Registered in the catalogue (spec chapter, ADR-012-consistent with the typl precedent) with LSP `codeDescription` targets reserved.

## Review

Per-task spec+quality reviews on all 13 plan tasks, plus a whole-branch final review. Its one Important finding — file-local false-positive `UXIL-018` — was fixed in `1d6ba5a` (scope filter, maintainer-approved); remaining minors triaged and deferred (helper-dedup + memoized pattern compilation noted for follow-up).

## Tests

`just build` full gate: **3980 passed / 0 failed** (unit + e2e), typecheck + lint clean, binary compiles. `deno fmt --check` and `dprint check` clean. New coverage: family gating (inert / designated / false-positive guard / upstream-inert / explicit-`Type:`), exact-position anchoring per surface form, per-code fixtures, and an e2e blackbox suite covering the gate and the bare-vs-file-local scope split.


stasson:
## /review findings

8 finder angles (3 correctness, 3 cleanup, altitude, conventions) over the full PR diff, each candidate independently verified against the branch. 18 raw candidates → 4 survived verification (1 refuted, the rest deduped, already tracked in #807, or consciously-accepted design decisions). Ranked most-severe first.

### 1. `compile()` and the MCP server never run the uxil family — surfaces disagree on the same corpus — CONFIRMED

`packages/markspec/core/compiler/mod.ts:290` calls the bare `validate(allEntries)`, which never invokes `validateUxilFamily`; only `runPipeline` (`check`) and LSP `validateAll` were wired. A designated project with a broken uxil declaration passes `markspec compile/export/show` and the MCP `validate` tool cleanly while `check` and the editor flag it — an agent or CI step trusting `compile`/MCP output treats a broken contract as clean. The asymmetry is real: `validate()` runs `validateTypl` internally, so TYPL codes ARE visible in `compile`, making the uxil gap surprising. `compile()` already holds `options.profile` (used at lines 294/300/316/336), so the wiring is a one-call addition; `mcp/project.ts:460` even documents its diagnostics as "CLI `check` parity" — that parity now has a hole. The e2e suite only exercises `check`, so this is untested.

**Suggested fix:** call `validateUxilFamily(allEntries, options.profile ?? null)` in `compile()`'s validation phase (cross-entry codes are safe there — compile always sees the full requested set), plus a compile-surface test.

### 2. `FILE_LOCAL_SUPPRESSED_UXIL` can silently drift from the code registry — CONFIRMED

`packages/markspec/core/validator/pipeline.ts:47-51` hardcodes `{UXIL-016, UXIL-017, UXIL-018}` as a plain `Set<string>`, 200 lines of orchestration away from `UXIL_CODES` (`core/uxil/diagnostics.ts:41-44`, whose `UxilCodeEntry` carries no scope metadata). The three codes are semantically principled (they fire from *absence* in a possibly-partial registry, unlike 019–022 which require presence), but nothing structural — no type, no test — keeps the set in sync: a future scope-sensitive UXIL code added to the registry reintroduces the file-local false-positive class this PR just fixed.

**Suggested fix:** add `scope: "cross-entry" | "entry-local"` to `UxilCodeEntry` and derive the filter from the registry (natural fit — the registry already centralizes severity/template per code).

### 3. LSP path compiles O(entries × types) regexes per debounced `validateAll` — CONFIRMED (modest today, scales with corpus)

`core/validator/uxil_family.ts:72` falls through to `classifyEntry` for every entry (the parser never sets `entry.type` — `core/parser/markdown.ts:533-536` — and the LSP never runs Stage 2), and `tryCompileDisplayIdPattern` (`core/validator/pattern.ts`) does a fresh `new RegExp()` per call with no memoization. Only fires when a profile designates a declaring type (the inert path short-circuits at `uxil_family.ts:64`). At ~300 entries × ~20 types this is ~6k throwaway regex compiles per keystroke debounce — single-digit-to-tens of ms, not perceptible today, but linear in corpus size.

**Suggested fix:** memoize compiled patterns per `(profile, pattern)` (e.g. `WeakMap<EffectiveProfile, Map<string, RegExp>>`), which also benefits every other `classifyEntry` caller.

### 4. 43-line `makeProfile` test helper copy-pasted byte-for-byte within this PR — CONFIRMED

`core/validator/uxil_family_test.ts:20-62` and `core/validator/pipeline_test.ts:30-72` are identical (`diff` returns empty), with a third inline variant in `lsp/workspace_test.ts:409-460`. Hand-rolled `EffectiveProfile` literals are a pre-existing repo-wide convention (17 test files), so this is inherited debt — but duplicating one's own new helper twice in a single PR was avoidable, and every future required `EffectiveProfile` field now touches 3 more sites. A shared test profile-builder would also serve S10 (#728).

### Refuted / already tracked

- "uxil anchoring helpers duplicate typl's `bridgeTyplDiagnostic`" — **refuted**: the typl bridge implements a materially different rule (unconditional column passthrough, fence-line arithmetic); a shared helper would not drop-in fit.
- typl's own cross-entry codes (TYPL-009/010/011) firing file-locally — pre-existing, already tracked as #807.

---

Findings 2–4 are non-blocking cleanups. Finding 1 is worth deciding before merge: either wire `compile()`/MCP now or record the surface split deliberately.

stasson:
**/review finding 1 addressed** in `407c770`: `compile()` now runs the uxil family after Phase 2.5 classification (`Phase 2.6`), so `markspec compile/export/show` and the MCP server agree with `check` and the LSP on designated corpora. No file-local filter in compile — the requested path set is the corpus (same full-set semantics as MSL-L006 there).

Coverage: 2 compiler unit tests (UXIL-009 present with a designated profile, no UXIL codes without) + 1 e2e (`markspec compile` on a bad-verb corpus → exit 0 with `error[UXIL-010]` on stderr, matching compile's pre-existing "diagnostics never flip the exit code" semantics). Full gate: 3983 passed / 0 failed. Findings 2–4 remain deferred cleanups.

